PR #18, opened 2013-10-29
Tolerate 0-value (missing) local size headers.

hempels:
Some rogue zip creators fail to set a non-zero local header size in ExtraData causing SharpZLib to reject them, this patch allows the library to tolerate that particular oversight like most other libraries.


superjulius:
Any plan or idea if this change will be merged. I am facing the same problem with some users and I would prefer to use a new release instead of recompiling the lib.

PR #49, opened 2014-05-12
Add support for reading original length of input stream

ajlennon:
I overrode Length in GzipInputStream to provide the original length of the compressed file, which is useful to me for my use case. Is this of interest?


McNeight:
Under evaluation.

-Neil

piksel:
I am closing  this due to lack of MIT and activity

PR #52, opened 2014-07-06
Adjust filename case.

hultqvist:
And added .gitignore for bin/obj

The name change is critical for the compilation to work on case sensitive file systems(in general Linux).


ermshiperete:
:+1:
This change is really necessary to be able to build on Linux.

McNeight:
This PR was obsoleted when PR #65 was merged into master at e01215507cf25a5978a0bd850c9e67dbabf515b7

-Neil

PR #55, opened 2014-08-13
Portable Class Library version

ygrenier:
Create solution and projet for the Portable version.
Create the conditionnal constant 'PCL' for code isolation.


Protagores:
Is this still relevant?

I would like to see a PCL version of this library which would be up to date. Preferably as `Profile136`, since it (apparently?) supports as many targets as possible while compatible with .NET 4.0.

Issue #54, opened 2014-08-11
Does not extract folders correctly in OSX (Mono)


augiem:
A zip file containing a single .bundle package (essentially a folder structure) when unzipped using Libraries.ICSharpCode.SharpZipLib.Zip.FastZip.ExtractZip extracts incorrectly in OSX, but works in Windows.

The resulting unzip extracts all files inside the .bundle package as files named for the folders they were in. For example, a .bundle containing 3 files which should be extracted to a single .bundle file will instead be extracted as follows:

name.bundle/Contents/MacOS/name
name.bundle/Contents/Resources/en.lproj/InfoPlist.strings
name.bundle/Contents/Info.plist

Keep in mind these are not folders being created -- its creating 3 files with the folder structure baked into their names.

I haven't tested a regular compressed folder, but .bundles are not being extracted correctly.

EDIT: Same problem happens with regular directories not in a bundle.

I've determined its because it's using "\" for the directory separator char instead of Path.DirectorySeparatorChar throughout due to the processing by the WindowsNameTransform class.

McNeight:
This issue was resolved when PR #65 was merged into master at e01215507cf25a5978a0bd850c9e67dbabf515b7

-Neil

PR #65, opened 2015-01-31
Fix handling of paths on Linux and Mac (issue #54)

ermshiperete:
Linux and Mac use a forward slash instead of a backslash to separate
directories. This change makes use of the Path.DirectorySeparatorChar
instead of hardcoding backslash. This allows to use the same code on
all platforms.

This change also adjusts the WindowsNameTransform tests so that they
also work cross-platform.

Incorporated in this PR is #52.


rversaw:
What's the status of this PR? Is it likely to be accepted soon?

odyth:
Bump.  Any update on this getting merged?

jfreilly:
WindowsNameTransform was never intended to be generic a clue being the name including Windows but this works fine.  It would be good to rename the class but this itself causes issues.

ermshiperete:
My Contribution is licensed under the MIT License.

PR #66, opened 2015-02-03
Allow the project file to be added to a VS 2013 solution

benbrown-x:
When attempting to use ICSharpCode.SharpZLib.csproj as part of a VS 2013 solution, an error occurs informing you that a one-way upgrade is required for compatibility.  Adding the ToolsVersion property to the project file allows the project to be added to more recent VS solutions in a way that shouldn't break existing builds or solutions.


benjamb:
#103: :+1:

McNeight:
Ben,

Thanks for responding, and thanks for your contribution.

-Neil

PR #70, opened 2015-02-20
Don't use SkipToNextEntry when checking Tar headers

ClemPi:
If two consecutives entries should be ignored the second one will pass


piksel:
Indeed. The way the tar headers are parsed is wrong.

piksel:

PR #71, opened 2015-02-20
Don't ignore Link entry type.

ClemPi:
There is no reason to ignore this kind of entry as TarHeader.LinkName is correctly filled. It's up to the user to handle it correctly.


piksel:
Fixed by #93.

PR #76, opened 2015-04-27
Closes #41 - Do not fail when local header compressed size equals to 0

### LINKED ISSUE #41: Ignore "Compressed size mismatch" error from Xceed
[SD-1801](http://bugtracker.sharpdevelop.net/issue/ViewIssue.aspx?id=1801), originally created on 12/20/2010 02:06:14 by David Pierson

When a password is applied, Xceed is incorrectly setting values in the
local header. This results in a “Compressed size mismatch between
central header(2399) and local header(0)” exception in
ZipFile.TestLocalHeader. 
 
 Sample code in ZipFile.cs at line 1259 ...
 
 if (compressedSize != entry.CompressedSize &&
                         compressedSize != 0xFFFFFFFF && compressedSize
!= -1**&& compressedSize != 0**) {
                         throw new ZipException(
                             string.Format("Compressed size mismatch
between central header({0}) and local header({1})",
                             entry.CompressedSize, compressedSize));
                     }
 
 Forum thread:

[http://community.sharpdevelop.net/forums/t/12404.aspx](http://community.sharpdevelop.net/forums/t/12404.aspx)
 "Errors unzipping zip files created using Xceed V4.0"

bbenoist:
As extracting some "valid" ZIP files was not possible...

Example ZIP file failing before the fix:
http://www.cmake.org/files/v3.2/cmake-3.2.2-win32-x86.zip

Also related to:
* https://github.com/icsharpcode/SharpZipLib/issues/41
* http://community.sharpdevelop.net/forums/t/12404.aspx

### DIFF
--- a/src/Zip/ZipFile.cs
+++ b/src/Zip/ZipFile.cs
@@ -1258,7 +1258,8 @@ namespace ICSharpCode.SharpZipLib.Zip
 					}
 
 					if (compressedSize != entry.CompressedSize &&
-						compressedSize != 0xFFFFFFFF && compressedSize != -1) {
+						compressedSize != 0xFFFFFFFF && compressedSize != -1 &&
+						compressedSize != 0) {
 						throw new ZipException(
 							string.Format("Compressed size mismatch between central header({0}) and local header({1})",
 							entry.CompressedSize, compressedSize));


bbenoist:
cc @jfreilly

bbenoist:
Duplicates #18, I did not see that these changes had been already proposed for merging.
Because the commits does not have the same content and I did not had enough time for improved testing, please consider merging #18 instead of #76.

McNeight:
I need to look at both #18 and #41 and compare with this issue to make sure everything lines up.

-Neil

McNeight:
PR #18 covered the exact same lines of code as PR #76, but was more complete.

Per originator request, PR #18 was merged instead of this PR.

-Neil

PR #83, opened 2015-07-13
Add new FastZip.ExtractZip overload with lambda file and directory fi…

depend86:
…lters


piksel:
I am closing this due to lack of response. If @depend86 gives permission to relicense it as MIT, we can open this again.

Issue #80, opened 2015-06-21
InflaterInputStream Length property is not set before fill()


kkguo:
InflaterInputStream.Length is 0 before executing read from it.

https://github.com/icsharpcode/SharpZipLib/blob/master/src/Zip/Compression/Streams/InflaterInputStream.cs#L552

This will cause following sample code fail, 
```c#
using (Stream strm = zipfile.GetInputStream(item))
{
          using (BinaryReader s = new BinaryReader(strm))
          {
                 Data = s.ReadBytes((int)s.BaseStream.Length);
           }
}
```
If CompressionMethod is Deflated, the data will be always empty as s.BaseStream.Length is 0.
probably use zipentry.size as the length.

BrisingrAerowing:
I did this in a slightly different way. I added a new constructor to the InflaterInputStream that passes the size as the second parameter and stores it in a new variable called size_, which is initialized to -1L. The Length property then checks if size_ is >= 0, and returns it if it is, otherwise it proceeds as in the original code. In the ZipFile class, I simply changed the construction of the stream to pass entries_[entryIndex].Size as the second parameter. This avoids an exception (and the possible breaking of other applications).

McNeight:
Fixed with pull request #84 

-Neil

PR #84, opened 2015-07-15
the stream is not seek-able, access to Length should return exception

kkguo:
related to issue #80 

just simply throw exception is enough, to get the real stream size before reading, should use entry.size.


McNeight:
Jingang,

-Neil

PR #85, opened 2015-09-01
Fix severe perf degradation reading Stored file in Mono

pldg:
When reading a stored file, ZipFile returns a PartialInputStream
directly instead of wrapping it with a decompression layer.

PartialInputStream handles sharing the underlying stream among
multiple threads by calling Seek() before each read.

Unlike Microsoft's reference implementation, Mono's FileStream
flushes its internal read buffer on every Seek.


McNeight:
Paul,

Thanks for responding, and thank you for your contribution.

-Neil

PR #86, opened 2015-09-16
Fix tar long names

creker:
Was using the library to generate debian packages which use Tar inside. Files with names longer than 100 symbols were preventing the package from installing - dpkg couldn't copy the files with the long names. Looking at the specification and Tars generated on Linux with default tools I found several bugs:
- Ustar format supports names up to 100 symbols. That is, if a name is exactly 100 symbols you don't need to write terminating symbol. SharpZipLib truncates names to 99 symbols
- "@LongLink" should have default 644 mode. SharpZipLib doesn't specify it
- "@LongLink" should have user ID, user name, group ID and group name from the entry it belongs to. SharpZipLib doesn't specify them

### DIFF
--- a/src/Tar/TarHeader.cs
+++ b/src/Tar/TarHeader.cs
@@ -896,7 +896,7 @@ namespace ICSharpCode.SharpZipLib.Tar
 
 			int i;
 			
-			for (i = 0 ; i < length - 1 && nameOffset + i < name.Length; ++i) {
+			for (i = 0 ; i < length && nameOffset + i < name.Length; ++i) {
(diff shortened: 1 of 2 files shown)


creker:
I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the #develop open source product (the "Contribution"). My Contribution is licensed under the MIT License.

McNeight:
Artem,

Спасибо за отвечать на запросы, и поблагодарить Вас за Ваш вклад.

-Neil

PR #92, opened 2015-11-11
Update ZipFile.cs

bhavin07:
AddUpdate method previously had index which was adjusted by -1 so, that it could with with updates_ generic list.
For some reason, -1 adjustment was removed in the latest code which was causing index out of bound error.

FindExistingUpdate method returns the one based index whereas updates_ generic list is zero based, so adjusting "index" by -1 is necessary.


McNeight:
Under evaluation.

-Neil

piksel:

PR #96, opened 2015-12-25
Bugfix/unicode Unicode File/Directory Support

aviranco:
Bugfix for Unicode support of file and directory names.
Previously created corrupted file/directory names, that underscore character ('_') took place instead of each unicode character in the file/directory name. e.g.:  _____.js


piksel:

PR #97, opened 2016-01-08
Allow disabling Multi-member gzip support for compatibility reasons

Greybird:
Allows circumventing regressions introduced by 5f464df0d4b01ea9ddc1c556ac2e25c388a0c3b1

I would like to propose you to introduce a way to deactivate multimember support as it created a regression on our application.
I totally understand that this is some kind of a workaround, but it could be of value in some specific case, and not having this would prevent migration.

Regards,

Arnaud


McNeight:
Under evaluation.

-Neil

Greybird:
Has been fixed with #117 for issue #116

Issue #91, opened 2015-11-08
Licensing


bjorn-ali-goransson:
I cannot find a clear explanation why the project is licensed GPL + exception and not either GPL or LGPL (or something else).

What's the problem with using a standard license?

rapgro:
sharpziplib-doc.noarch: E: incorrect-fsf-address
/usr/share/doc/sharpziplib-doc/COPYING.txt
sharpziplib-doc.noarch: E: incorrect-fsf-address
/usr/share/licenses/sharpziplib-doc/COPYING.txt

Please use a current license text.
https://gnu.org/licenses/gpl.html

McNeight:
The nice thing about standards is that there are so many to choose from, and that includes the GPL.

http://www.gnu.org/software/classpath/license.html

-Neil

jfreilly:
The not nice thing about open source is that it can be very hard to change licenses.  A whole lot of people have signed  the existing license.  What happens when you can't easily get in contact with all those people any more?   Easy thing is to leave the license alone.

rapgro:
What? My point is about wrong / old license text, there's a new address of FSF.

Issue #89, opened 2015-10-08
incorrect fsf address


rapgro:
The Free Software Foundation address in the doc/COPYING.txt file seems to be outdated or
misspelled.
Current license text can be found here:
http://www.gnu.org/licenses/old-licenses/gpl-2.0.html

rapgro:
Files with incorrect fsf address, found with licensecheck scrpt:

SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/AssemblyInfo.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/BZip2/BZip2.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/BZip2/BZip2Constants.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/BZip2/BZip2Exception.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/BZip2/BZip2InputStream.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/BZip2/BZip2OutputStream.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Checksums/Adler32.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Checksums/Crc32.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Checksums/IChecksum.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Checksums/StrangeCrc.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Core/FileSystemScanner.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Core/INameTransform.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Core/IScanFilter.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Core/NameFilter.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Core/PathFilter.cs
SharpZipLib-e01215507cf25a5978a0bd850c9e67dbabf515b7/src/Core/StreamUtils.cs
[...]

PR #101, opened 2016-04-01
Copyright & Licensing

McNeight:
* Update address of Free Software Foundation, which closes icsharpcode/SharpZipLib#91
* All of the various copyright statements have been consolidated to the following:
Copyright © 2000-2016 AlphaSierraPapa for the SharpZipLib Team

### DIFF
--- a/doc/COPYING.txt
+++ b/doc/COPYING.txt
@@ -2,7 +2,7 @@
 		       Version 2, June 1991
 
  Copyright (C) 1989, 1991 Free Software Foundation, Inc.
-                       59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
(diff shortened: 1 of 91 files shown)


McNeight:
This also closes icsharpcode/SharpZipLib#89, and in case it is necessary:

_I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the #ziplib open source product (the "Contribution"). My Contribution is licensed under the MIT License._

christophwille:
How about we consolidate a tad more - a single license file, no license header/author headers any more. We did that with other projects too.

McNeight:
Since some of the files were adapted from the GNU Classpath library, do we want to have the entire library covered by that one license? Or do we want to have everything MIT licensed except for those files?

christophwille:
The origins of SharpZipLib... it started out its life as a quick port of Java code (@mkrueger did the original version), and out of respect for this, the identical license was adopted. The origins show insofar as the naming is inconsistent with the .NET naming conventions. (it was originally used in SharpDevelop to compress code completion data, that's why we did in the first place)

We meant to change this long ago (meaning tidy the code), but it never happened. The reason to consider doing this at all was mostly the many people asking about the "weird" license. Yes, having all MIT would be way better contribution-wise as well as end-user wise.

jfreilly:
Looks fine

xied75:
Regarding the Java legacy, is that still a plan to rewrite/remove those? Given it was done long time ago, (at the time Java was lacking lots of today's modern features, let alone .NET), I feel it should be worth it to simply rewrite all the Java files. If you guys agree we may create an Issue to cover those tasks.

PR #127, opened 2016-07-28
Fixed 10 Tests (out of 16 Failing)

hypersw:
Would like to try optimizations on stream read virtual calls, for which I need to check the changes with tests. Turned out that 16 of them were failing out of the box. Fixed 10 which had more or less obvious technical issues. The idea of the remaining tests is not so straightforward.
This applies to WinNT platform.


McNeight:
Serge,

-Neil

PR #129, opened 2016-08-05
Added a CompressionLevel enum to Deflater to allow FastZip users to s…

zippo227:
I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the SharpZipLib open source product (the "Contribution"). My Contribution is licensed under the MIT License.

I came across this feature need in the forum because I also needed it.
[http://community.sharpdevelop.net/forums/p/6835/19472.aspx](http://community.sharpdevelop.net/forums/p/6835/19472.aspx)

I recently switched from DotNetZip because it was not functioning well with my project when unzipping it in OSX. SharpZipLib did not have the same problem, but I did miss the compression level feature. I followed the practices I saw in the library, and I added the enum to Deflater because that is where the const int definitions were. I personally think it would have been cleaner to add my enum directly to the ICSharpCode.SharpZipLib.Zip.Compression namespace, but I leave that up to you.

### DIFF
--- a/ICSharpCode.SharpZipLib/Zip/Compression/Deflater.cs
+++ b/ICSharpCode.SharpZipLib/Zip/Compression/Deflater.cs
@@ -77,9 +77,45 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression
 		/// There is no need to use this constant at all.
 		/// </summary>
 		public const int DEFLATED = 8;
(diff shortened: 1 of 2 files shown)


zippo227:
I'm glad you added this to the milestones! @McNeight

PR #133, opened 2016-08-14
PendingBuffer ToByteArray() to return all bytes

charlesgriffiths:

If AlignToByte() (or a procedure similar to that in Flush()) is not done within ToByteArray(), the bits variable may contain up to 15 bits not present in buffer_ which could result in data truncation. I have added AlignToByte() on the assumption that ToByteArray() will only be called when all output to PendingBuffer is complete.


McNeight:
Charles,

-Neil

PR #134, opened 2016-08-16
change while to if

charlesgriffiths:

This while should never loop


McNeight:
@charlesgriffiths 

-Neil

PR #135, opened 2016-08-16
FindLongestMatch looks beyond lookahead

charlesgriffiths:

If lookahead is less than MAX_MATCH, FindLongestMatch can look beyond where it should. This is not a serious problem because matchLen is checked before returning, but in a rare case I believe a false match could obscure a better match before the former is found out.

Testing shows the new function runs slightly slower than the old version, but the difference is generally less than 1% of total runtime.


McNeight:
@charlesgriffiths 

-Neil

PR #138, opened 2016-09-11
Implemented NETStandard1.3 lib and made changes get lib compiling

zone117x:

Creating a pull request for review purposes. These changes need reviewed and tested before merging. 

No features were removed during this refactoring other than some exception serialization related stuff. 
- The ZIP AES changes need tested
- The streams were improperly overriding Stream.Close() according to https://msdn.microsoft.com/en-us/library/ms143450(v=vs.110).aspx ... these were changed to Stream.Dispose(bool) but there may be problems around garbage collection - needs reviewed and tested
- Unsupported exception serialization and constructors were removed
- DefaultCodePage changes need reviewed
- Some commenting may need updated

### DIFF
--- a/ICSharpCode.SharpZipLib.sln
+++ b/ICSharpCode.SharpZipLib.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ICSharpCode.SharpZipLib", "ICSharpCode.SharpZipLib\ICSharpCode.SharpZipLib.csproj", "{0E7413FF-EB9E-4714-ACF2-BE3A6A7B2FFD}"
 EndProject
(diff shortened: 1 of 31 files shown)


xied75:
Agree on the change to remove using Stream.Close(), it's no longer exist in the coreclr/corefx.

zone117x:
Separating them is a bit hard because the above changes were the minimum required to get the .netstandard lib to compile at all. 

We could make those individual changes only for the existing lib with netstandard and PR those one at a time, but I don't have time to do that. I got my fork working for my purposes and mostly made this PR to give some head start to the main contributors.

xied75:
If @McNeight is willing to consider merge it, I can volunteer to help the review.

jahmai:
Missed specifying code signing key on the netstandard project.
Also, nuspec not updated.

Otherwise, I have built and modified the existing nuget package by hand and our unit tests that use this still pass, so well done 👍

McNeight:
This is great work! My lingering concern, and the reason why I took so long to finally commit this, is that it breaks some of the automated build and test. I've been hoping to get it cleaned up before now, but I've realized that I need to get the change in and then work on cleaning up.

Any help with getting the automated build and test setup back to normal again would be much appreciated.

-Neil

xied75:
I did a clean clone and trying to build with VS2015up3, somehow I got this error:
```
1>C:\Program Files (x86)\MSBuild\Microsoft\NuGet\Microsoft.NuGet.targets(140,5): error : Your project is not referencing the ".NETFramework,Version=v4.5" framework. Add a reference to ".NETFramework,Version=v4.5" in the "frameworks" section of your project.json, and then re-run NuGet restore.
```

In fact I could right click and just build the portable project, but not the main 4.5 project.

Just wondering, netstandard1.3 will support from 4.6, are we going to keep 4.5 project?

zone117x:
This is fixed in my last pull request https://github.com/icsharpcode/SharpZipLib/pull/150

PR #145, opened 2016-10-15
Efficient skip in TarInputStream

adv12:

Added a small bit of code that allows a TarInputStream to efficiently
skip portions of its underlying InputStream without screwing up the
bookkeeping in TarBuffer.

### DIFF
--- a/ICSharpCode.SharpZipLib/Tar/TarBuffer.cs
+++ b/ICSharpCode.SharpZipLib/Tar/TarBuffer.cs
@@ -145,7 +145,11 @@ namespace ICSharpCode.SharpZipLib.Tar
 
 			var tarBuffer = new TarBuffer();
 			tarBuffer.inputStream = inputStream;
-			tarBuffer.outputStream = null;
+            if (inputStream.CanSeek)
+            {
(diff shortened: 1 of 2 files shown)


piksel:
I realize that this is pretty old and cannot be merged as is anyway, but this really needs tests to "prove" that it can safely skip the parts. If it gets updated and some tests are added I can reopen this.

adv12:
Thanks for the response.  I no longer work for the company that needed this fix and don't have the surrounding code or good example data, but if I ever get access to all that again, I'll try to submit something with tests.  Incidentally, this is 1/2 of the equation for being able to seek quickly around TAR archives; the other half is fixing TarInputStream.Position, which does not accurately report the true position in the stream.  I submitted a pull request for that as well, but it was closed as wontfix.

piksel:
Yeah, sorry about that. But both of them had been stale for several years without anyone touching them, and as I wrote then, I didn't see how that was supposed to be used. Going through and verifying changes like this really needs to motivated by an outstanding issue etc, otherwise it's just adding risk and maintenance burden to a project that already has limited resources as is.

PR #146, opened 2016-10-15
TarInputStream.RealPosition

adv12:

Added a property called RealPosition that returns the position within
the stream of the next byte that will be read by Read(), accounting for
the internal buffering that makes the Position property useless.

### DIFF
--- a/ICSharpCode.SharpZipLib/Tar/TarInputStream.cs
+++ b/ICSharpCode.SharpZipLib/Tar/TarInputStream.cs
@@ -97,10 +97,33 @@ namespace ICSharpCode.SharpZipLib.Tar
 			}
 		}
 
-		/// <summary>
-		/// Flushes the baseInputStream
-		/// </summary>
-		public override void Flush()
+        /// <summary>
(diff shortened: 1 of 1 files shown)


adv12:
I was going to try to rebase this pull request after the reorganization of the code, but I don't have VS 2017 here at work so I can no longer open these projects.  Maybe I'll try from home with VS 2017 Community.  In case someone else tries to manually edit this change into the current master, a warning: the math currently overflows for large files because the properties used in the calculation are all ints.  They need to be cast to longs before being multiplied and added.  As I say, I may try to update this pull request from home, but I'm also pretty new to Git and may have trouble with the rebase.

piksel:
I will close this since we don't have a clear usage for this,  and I don't want to add/support additional features "just because".

adv12:
The use case is identical to the use case for Position on any stream.  The problem is that Position is implemented incorrectly for TarInputStream.  This was a workaround for that.  The proper thing to do would be to replace the implementation of Position with this implementation, but I submitted this as a workaround because I assumed by now people may be depending on the broken behavior.

PR #149, opened 2016-10-31
Migration to PCL using xproj + project.json, targetting .NET 4.5 and .NET Standard 1.3

gingermike:
As part of some work I've recently undertaken in a project to migrate over to .NET Core 1.0 on Linux, I've had to migrate SharpZipLib to a 'proper' .NET Standard library. In the process of this, I've moved the project definitions over to use xproj.

All tests run with the same results on the master branch (there are currently two failing for me?). I have to run them on the command line using the dotnet core runtime (run 'dotnet test' in the root of the test project).

My consuming project only uses the zip functionality of the library, which appears to work fine post migration.

I've not had a chance to migrate the sample projects over to use the new libraries.


zone117x:
"Migration to PCL using xproj + project.json" is a bit of a nonsensical statement - see Microsoft's blog posts on NET Standard where they explain the new architectures (Netstandard is a replacement for PCL):  https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/

Microsoft changed course with both xproj and project.json file types back in May https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/
And is in the process of moving all the new stuff back into csproj files, and getting all the csproj tooling like msbuild working cross platform. 

The current master branch with the netstandard.csproj is much closer to what the finalized netstandard project type will look like. 

Adding `#if NET45` flags all over and xproj/project.json should not be necessary for this lib given the direction Microsoft is going.

HarelM:
VS 2017 now supports the new format, any plans for migration/support?

PR #150, opened 2016-11-02
Fix compiling of original NET45 csproj, better default encoding, ...

zone117x:

* Side-by-side csproj files don't work when the folder contains a project.json. Visual Studio attempts to use the project.json when attempt to compile the regular .net45 lib. I moved the csprojs into separate folders to fix this.

* There was also an issue with NET45 not including the IncrementalHash that replaced the legacy HMACSHA1 class - so I wrote a shim for for use with NET45.

* The default codepage encoding code was not good. I changed it from defaulting to system code page to now defaulting to UTF-8. This change is inline with the original ZIP specification and supports internalization better. 

* There were a few non-generic collection types being used which for NETStandard required a nuget dependency on "System.Collections.NonGeneric" - so I changed them to the generic  types. 

The tests project runs and all but 2 pass. The 2 failures seem unrelated to recent changes (something with datetime and other with file locking).


McNeight:
@zone117x 

Thanks again for your contribution.

-Neil

Issue #103, opened 2016-04-05
MIT License?


xied75:
Dear Team,

Given the whole .NET world is now MIT, any chance this great lib could also change to that?

christophwille:
This is currently also discussed in pull request https://github.com/icsharpcode/SharpZipLib/pull/101

McNeight:
The following PRs were merged without MIT license assignment:
#1 by @linquize (2 lines in 1 file)
#4 by @Mailaender (2 lines in 1 file, and 2 deleted files)
#7 by @Mailaender (8 lines in one new file)
#10 by @RobThree (1 line in 1 file)
#53 by @novn2013 (1 line in 1 file)
#64 by @hypersw (10 lines in 1 file)
#65 by @ermshiperete and @hultqvist (16 lines in 6 files)
#66 by @benjamb (1 line in 1 file)

hultqvist:
Do the listed PR contain enough to qualify for copyright and thus need the assignments. Looking at my own contribution (i changed the case in a few filenames) can't possibly reach verkshöjd(can't bother to translate) to be covered by copyright.

Although if you want me to take any legal action I will happily contact a lawyer and pass on the bill to you.

McNeight:
Peter,

Inget behov av rättsliga åtgärder. Tack för att svara, och tack för ditt bidrag.

-Neil

christophwille:
Good morning,

I am the "all things administrative red tape guy" on the SharpDevelop team. Some background on why we now want to (finally) take SharpZipLib to MIT - and action items.

@McNeight has joined as the new maintainer for SharpZipLib, and in the process of discussing with the old maintainers @jfreilly and @davidpierson somewhat obviously the license topic came up (and I can tell everybody that it was/is a pain to me too just by looking at the inbox on how many questions I got over the years just on license clarification).

Now, SharpZipLib started out of part of SharpDevelop, and with SharpDevelop we required JCAs (please see https://github.com/icsharpcode/SharpDevelop/wiki/Joining-the-Team). We want to get to exactly the same point with SharpZipLib, and so we want to go the following route:

 * The maintainer (@McNeight in this case) has a JCA on file. Check.
 * All PRs (that qualify as a patch) have the MIT/assignment paragraph from https://github.com/icsharpcode/SharpDevelop/wiki/Joining-the-Team as part of it. This issue is the thread for it.
 * The original port is cleaned up (which is necessary technically anyways)

[...]

PR #169, opened 2017-03-19
Update index.html

yinyue200:
Update license info


christophwille:
Shouldn't we directly link to #103? (not only textual mention)

I'd recommend merging this once we actually have a nuget package up that is sporting the changed license.

PR #179, opened 2017-05-01
inline crc32 checksum update, about 2x higher crc throughput

dpethes:
Removes redundant function call for each checked byte. The speedup is most noticeable when decoding highly compressed data, where crc takes a significant amount of time.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Checksum/Crc32.cs
+++ b/src/ICSharpCode.SharpZipLib/Checksum/Crc32.cs
@@ -181,9 +181,13 @@ namespace ICSharpCode.SharpZipLib.Checksum
 				throw new ArgumentOutOfRangeException(nameof(count), "exceeds buffer size");
 			}
 
-			for (int i = 0; i < count; ++i) {
-				Update(buffer[offset++]);
-			}
-		}
+            uint crc = checkValue;
+            for (int i = 0; i < count; ++i)
+            {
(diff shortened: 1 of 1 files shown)


christophwille:
Is there a specific reason for removing unchecked() ? Original code in Update(...):
checkValue = unchecked(crcTable[(checkValue ^ bval) & 0xFF] ^ (checkValue >> 8));

dpethes:
It has no measurable effect performance-wise.

christophwille:
well, I think it could have an effect exception-wise (wasn't worrying much about perf):
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/unchecked

christophwille:
"2x higher throughput" - did you measure that with a Release build or where does this number come from? Because I can't imagine that the runtime wouldn't inline this method call.

dpethes:
The crc calculation is doing only bitops, unchecked matters only for arithmetic ops.
IIRC I modified one of the apps in the Samples solution* for the measurement. I used VS2015 release build, but I don't remember if it was run as 32bit or 64bit - most likely latter. Inlining the method isn't enough - it would have to skip updating the checksum field during calculation as well.
If you have some samples for perf testing that you prefer, I could prepare some new measurements.


* I had to fix some references there - is it still broken?

piksel:
See #306

Issue #121, opened 2016-07-06
TAR in ustar format with long file names: TarEntry name is incomplete


alexmbaker:
### Steps to reproduce
1. Obtain a TAR with long file names - eg http://registry.npmjs.org/npm/-/npm-3.10.3.tgz
2. Follow the directions to extract from the archive provided in the project WiKi  https://github.com/icsharpcode/SharpZipLib/wiki/GZip-and-Tar-Samples#-simple-full-extract-from-a-tgz-targz 
3. Run the extract
### Expected behavior

The files should be extracted in the same folder structure as it would be using 7Zip.  The extracted should look something like

```
.-- package
     |-- bin
     +-- node_modules
     |    |-- abbrev
     |    |--   
     |    +-- read-package-json
     |    +-- etc etc
```
### Actual behavior

Where the file names including the relative path is long the files are extracted in the wrong place

```
.-- package
|    |-- bin
|    +-- node_modules
|    |    |-- abbrev
|    |    |--    
|
|-- read-package-json
|-- etc etc
```
### Additional Information

The code in `TarInputStream.GetNextEntry` appears to only try to read the additional information if the  header has a `typeFlag` equivelent to `L`.  Looking at this documentation https://www.gnu.org/software/tar/manual/html_chapter/tar_14.html it appears different versions of the format do different things.  The implementation in SharpZipLib appears to follow the 'old gnu'  way but does have support for the new / current way which appears to be to read the field and see what is there.
### Version of SharpZipLib

SharpZipLib.0.86.0 
### Obtained from (place an x between the brackets for all that apply)
- [ ] Compiled from source
  - branch: _______
  - commit: _______
- [ ] Downloaded DLL from GitHub
- [ ] Downloaded DLL from SourceForge
- [ ] Downloaded DLL from _______
- [ ] DLL included as part of
- Package installed using:
  - [x] NuGet
  - [ ] MyGet
  - [ ] Chocolatey

siegfriedpammer:
I tried reproducing this bug. I extracted npm-3.10.3.tgz using 7zip and TarArchive.ExtractContents and both directory structures are identical. Can you provide a dump of the directory structure you are getting? What code are you using to extract the files?

PR #182, opened 2017-07-04
writing correct Tar UTF8 filenames

shu66y:
We had the problem of writing UTF8 Filenames within a Tar correctly. I didn't consider reading so far.


piksel:
I have been going through how the tar headers are parsed for #121, and UTF-8 encoded names is probably better. The specification suggests using extended headers for encodings other than ASCII, but GNU Tar writes the UTF8-encoded bytes as the `Name` so I guess we should too.

src/ICSharpCode.SharpZipLib/Tar/TarHeader.cs
```diff
@@ -870,9 +870,11 @@ public static int GetNameBytes(string name, int nameOffset, byte[] buffer, int b
 
 			int i;
 
-			for (i = 0; i < length && nameOffset + i < name.Length; ++i)
+			byte[] nameBytes = Encoding.UTF8.GetBytes(name);
```
piksel: I doubt we want BOMs in the name

```suggestion
			byte[] nameBytes = new Encoding.UTF8Encoding(false).GetBytes(name);
```

itn3000:
This PR's may cut the filename which has long length(at least 100 bytes in UTF-8) + non-ascii character because char above U+80 is not one byte in UTF-8.

Numpsy:
Is this obsoleted by #364 being merged?

piksel:
Indeed. Thanks.

Issue #187, opened 2017-08-24
State of the Union August 2017


christophwille:
You will have noticed the recent activity by @siegfriedpammer and myself. Neither of us is the "maintainer" of this repository: I am the "custodian" (best description I could come up with) of the ic#code projects, and I roped in @siegfriedpammer (ILSpy) to help me put finishing touches on the almost-finished netstandard1.3 work after the Visual Studio 2017 Version 15.3 launch, which brought a couple of things:

 * .NET Core 2.0 final (which makes the unit tests work with NUnit properly)
 * NuGet support in VS 2015 (NuGet 3.6+ from nuget.org/downloads) to allow proper referencing of netstandard1.3 projects

This is reflected in that there are now -alpha builds available on NuGet.org for SharpZipLib. Why -alpha? Neither of us has followed the project for the past few years closely enough to make a proper judgement on the current quality at this point in time (we are still learning).

## What do you need to use the -alpha bits?

At least VS2015 with this applied https://blog.nuget.org/20170815/Whats-nu-in-NuGet-with-VS2017-15-3.html#update-to-the-nuget-client-in-visual-studio-2015

## Which .NET platforms will it work on?

See https://docs.microsoft.com/en-us/dotnet/standard/net-standard With -alpha1 targetting netstandard1.3, full .NET Framework 4.6 and newer for example. Details are in the matrix. Yes, this means for older projects, you are stuck with 0.86.

## About the license - what's the story here?

Yes, with -alpha1 the license switch to MIT is now fully realized. No more "It is amost like LGPL". Clean and simple.

## What do you need to open SharpZipLib's source code (dev)?

At least VS2017 Version 15.3 with the .NET Core 2.0 SDK installed.

## What about the open PRs with "NeedMIT" tag?

Those miss the following paragraph: "I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the SharpZipLib open source product (the "Contribution"). My Contribution is licensed under the MIT License."

Please always supply this. See Contributing.md.

## Want to help? (user of SharpZipLib)

Our pri1 item - please put the new -alpha packages through their paces! (notable change: Dispose instead of Close on streams)

## Want to help? (contributing to SharpZipLib)

[...]

PR #198, opened 2017-10-23
Restore support for NET4.5

zone117x:
The original target was only netstandard1.3 which is only supported on the full net framework 4.6 or higher.
The new csproj format allows easy multi-framework support so it now builds for netstandard1.3 and net45. 

### DIFF
--- a/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
+++ b/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -23,5 +23,12 @@ Please see https://github.com/icsharpcode/SharpZipLib/wiki/Release-1.0 for more 
     <DocumentationFile>bin\Release\netstandard1.3\ICSharpCode.SharpZipLib.xml</DocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net45|AnyCPU'">
+    <DocumentationFile>bin\Debug\net45\ICSharpCode.SharpZipLib.xml</DocumentationFile>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|net45|AnyCPU'">
+    <DocumentationFile>bin\Release\net45\ICSharpCode.SharpZipLib.xml</DocumentationFile>
+  </PropertyGroup>
 
 </Project>
--- a/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
+++ b/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
     <PackageId>SharpZipLib </PackageId>
     <SignAssembly>True</SignAssembly>
     <AssemblyOriginatorKeyFile>ICSharpCode.SharpZipLib.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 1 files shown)


christophwille:
Technically, that is correct (multiple TFMs). However, 1.3 was chosen intentionally, and according to https://docs.microsoft.com/en-us/dotnet/standard/net-standard it is *not* compatible with 4.5. The minimum supported version is intentionally 4.6.

zone117x:
Are you saying this project has intentially limited support to netframework46 / netstandard1.3 or higher? That would be a poor decision for a large lib given how many projects are stuck on netframework4.5. Especially since there is nothing in this repo that requires extra work to support netframework4.5, as this pull request demonstrates.

zone117x:
This includes WPF/winform devs targeting the latest net framework included on windows 7 (netframework4.5.1), and devs targeting the "Full" framework on Xamarin.Mac which is essentially netframework4.5.

christophwille:
Yes, this was an intentional decision, because we didn't want to be limited by having to support anything older (and at some point add #ifs).

piksel:
I don't think there should be much of a problem supporting both. No #if's should be needed.
Couldn't we remove net45 support later if any larger incompabilities come up?
The code compiles fine with both frameworks and I can take responsibilty for making sure it works and deal with any future compability issues.

zone117x:
Sounds like premature "future-proofing" that has little gain yet large downside for the massive number of LOB (line of business) developers. 

If you are really that concerned with ability to use newer APIs then might as well bump the minimum to netstandard2.0 since that's where most the major new features are.

zone117x:
And for that matter, many of the convienent newer features like Span<T> overloads in the core runtime exist only in netcore2.1, and the current official suggestion to lib maintainers to manage their usage with #ifs.

christophwille:
We had that very issue on another project - how would you feel to get net45 now to only have it removed for good in two months?

zone117x:
This lib is the defacto standard archive lib in .net - similar to the json.net lib. Do you not see how short-sighted it is dropping support for so many users?

christophwille:
We usually strive to align with OS support and the framework version installed there. And yes, if it were up to me, it would be netstandard2 because it is much better aligned with 4.6.2 and beyond (and much less hassle in restore scenarios).

zone117x:
Then hundreds of thousands of devs doing practical development with constraints in the real world would be happy if it was not up to you :)

piksel:
The problem right now is that we have so many people using the 0.86-version of the library that simply cannot upgrade to v1+.

christophwille:
This is because there are breaking changes in there. That's why it is a major version number. Hadn't the project been without maintainers for years we'd have had breaking changes a long time ago.

zone117x:
And if the main lib just drops them then it leads to additional fragmentation with devs continuing to use various forked nuget packages.

christophwille:
Which is the status quo anyways

zone117x:
There is a difference between breaking changes and inability to install the package.

zone117x:
The status quo for poorly maintained packages, yes.

christophwille:
The commit log is testament to this

zone117x:
Not sure what you mean. I bet most typical usage of this lib would not result in breaking changes if devs had that ability to update to the package.

christophwille:
I meant "take a look at the commit log of SharpZipLib, do you think it is well maintained"?

zone117x:
So your logic is that because it was not well maintained, it should continue to be poorly maintained by dropping support for users?

christophwille:
Dropping unmaintained (EOL'd) runtimes is different from the other.

zone117x:
I eagerly await your pull request to the newly released [System.Memory](https://www.nuget.org/packages/System.Memory/) package to drop support for netframework4.5. And to [Newtonsoft.Json](https://www.nuget.org/packages/newtonsoft.json/) to drop netframrwork 2.0 through 4.5.

christophwille:
To flip that around - you are signing up as maintainer of SharpZipLib for a couple of years and decide to volunteer let's say 5+ hours / week?

zone117x:
What a ridiculous question. A fairly stable lib like this requiring 5+ / hours a week per maintainer? 
I would help @piksel to do the minimal maintaining this project needs for netframrwork 4.5 support.

zone117x:
Ironically the only reply in the maintenance issue is regarding net45 support https://github.com/icsharpcode/SharpZipLib/issues/187

christophwille:
It is always the majority that remains silent

piksel:
I noticed that there is one place where #if NET45 is needed:
[src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs#L12](https://github.com/icsharpcode/SharpZipLib/blob/be53259f188e39c0965dfe8d36448ddbd706403b/src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs#L12)
Because of IncrementalHash which came (as an nuget package) in NET46. 
Still, I would not consider NET45 deprecated as long as there is a large user base.
As being pointed out above; the library serves some core functionality for many projects.

How many developers would be fine with shipping an installer that doesn't work on windows 7 for example? It still has a pretty large market share, and especially in corporate environments.

christophwille:
Windows 7 SP1 is a supported operating system (side note: that is roughly the take we always strive for - if a system still gets security updates, we try to also support it unless it is a major hassle - and if you have Windows 7 SP1 with security updates, you are on a way newer netfw version than 4.5).

zone117x:
> if you have Windows 7 SP1 with security updates, you are on a way newer netfw version than 4.5

Absolutely not the case. We have a technical product with a userbase that we report latest installed netframework for bug analysis. A huge number of users do not install anything past the bundled version.

zone117x:
> It is always the majority that remains silent

The LOB devs who need something like net45 support are by far the "most silent" on github.

PR #201, opened 2017-11-27
FastZip: Raise ProcessDirectory event for extractions

Stevie-O:
FastZip's ExtractZip method did not raise the ProcessDirectory event when
creating a directory.  There was no way to determine or control when
ExtractZip decided to create a new directory.

(This presented an issue when I tried using ExtractZip in conjunction as part of an installation tool with automatic-rollback functionality; it did not rollback the creation of directories, so the cleanup didn't work correctly.)

NOTE: The logic to raise the event only triggers when the target directory
doesn't already exist in the destination.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
@@ -783,7 +783,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 
 			// TODO: Fire delegate/throw exception were compression method not supported, or name is invalid?
 
-			string dirName = null;
+			string dirName = string.Empty;
 
 			if (doExtraction)
 			{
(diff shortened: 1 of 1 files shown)


src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
```diff
@@ -602,7 +602,13 @@ void ExtractEntry(ZipEntry entry)
 			if (doExtraction && !Directory.Exists(dirName)) {
 				if (!entry.IsDirectory || CreateEmptyDirectories) {
 					try {
-						Directory.CreateDirectory(dirName);
+						if (continueRunning_ = events_.OnProcessDirectory(dirName, true)) {
```
piksel: `events_` is not checked for null. This is probably what causes the tests to fail.

Could be solved with:
```.cs
if (continueRunning_ = events_?.OnProcessDirectory(dirName, true) ?? true) {
```

piksel:
I commited my suggestion and rebased on master, this could be useful for determining if directories were created upon extraction (as per the original description).

PR #205, opened 2017-12-14
DateTime of a ZipEntry when updating a ZipFile

mdima:
I just run in this same limit: http://community.sharpdevelop.net/forums/t/19039.aspx , so I added a new method.
Tested and working on my project.

Thanks,
Mdm

PS: I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the SharpZipLib open source product (the "Contribution"). My Contribution is licensed under the MIT License.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -1518,6 +1518,34 @@ namespace ICSharpCode.SharpZipLib.Zip
 
 
 		/// <summary>
+		/// Add a file to the archive.
+		/// </summary>
+		/// <param name="fileName">The name of the file to add.</param>
+		/// <param name="entryName">The name to use for the <see cref="ZipEntry"/> on the Zip file created.</param>
+		/// <param name="entryDateTime">The DateTime to use for the <see cref="ZipEntry"/> on the Zip file created.</param>
+		/// <exception cref="ArgumentNullException">Argument supplied is null.</exception>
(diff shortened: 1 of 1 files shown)


src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -1517,6 +1517,34 @@ public void Add(string fileName, string entryName)
...
+
+			if (entryName == null)
+			{
+				throw new ArgumentNullException(nameof(entryName));
+			}
+
+			CheckUpdating();
+			contentsEdited_ = true;
+
+			ZipEntry entry = EntryFactory.MakeFileEntry(entryName);
```
piksel: `ZipEntryFactory.MakeFileEntry(string fileName)` is an overload of `MakeFileEntry(fileName, null, true)` which means that this will try to read the file `entryName` and create an entry from that file.

If you change this to `ZipEntry entry = EntryFactory.MakeFileEntry(fileName, entryName, true);` it will work as intended.

I am not too sure about how useful this addition is, though. Adding an entry with a custom `DateTime` can be done through `ZipFile.Add(ZipEntry)`.

piksel:
Closing this in favor of #350 since it enables this use case by adding a overload to `ZipFile.Add` which takes a ZipEntry and a data source.
Usage:
```cs
var zipEntry = zipFile.EntryFactory.MakeFileEntry(fileName);
zipEntry.DateTime = entryDateTime;
zipFile.Add(new StaticDiskDataSource(fileName), zipEntry);
```

Issue #164, opened 2017-01-19
System.NotSupportedException when zipping more than ~4.2 gb of data to a stream doesn't support seeking


zlangner:
### Steps to reproduce
You have to zip up enough data so that when the stream is writing the end of the central directory the function ZipHelperStream.WriteZip64EndOfCentralDirectory will be called. The first thing this does is call stream_.Position which immediately fails if the stream being written to doesn't support seeking and therefore doesn't support position. [MSDN](https://msdn.microsoft.com/en-us/library/system.io.stream.position(v=vs.110).aspx) describes the relationship between the stream position property and the CanSeek property. 

### Expected behavior
Ideally, the stream's position would not be needed at all. Instead I think this should be tracked like the sizeEntries or startOfCentralDirectory is already being tracked by SharpZipLib.

### Actual behavior
Since the underlying stream does not support seeking the stream's position property throws a System.NotSupportedException which then bubbles up.

### Version of SharpZipLib
0.86.0.518

### Obtained from (place an x between the brackets for all that apply)
- [ ] Compiled from source
 - branch: _______
 - commit: _______
- [ ] Downloaded DLL from GitHub
- [ ] Downloaded DLL from SourceForge
- [ ] Downloaded DLL from _______
- [ ] DLL included as part of
- Package installed using:
 - [X] NuGet
 - [ ] MyGet
 - [ ] Chocolatey

insylogo:
Here's a PR that addresses this issue: https://github.com/icsharpcode/SharpZipLib/pull/226/

PR #226, opened 2018-04-25
Fixes issue #164 by not accessing stream position to determine central…

insylogo:
… signature offset (instead, use central directory offset + size of all entries)


piksel:
It certainly SEEMS like the end of the record could be determined by adding the start of the record and the size. I don't know the spec well enough to know if this is always the case tho.

insylogo:
You can verify that by saving a 64 bit zip (file stream not http) and observe the stream position is the same as this calculated position. I also verified that large streamed zips work after this change, but that first part would provide firm confirmation for you

PR #234, opened 2018-06-16
Update framework targets to netstandard2 and net45

piksel:
Update framework targets to netstandard2 and net45


christophwille:
ok (do we also later bring back Close semantics?)

piksel:
@christophwille I'm not sure what you mean with the close semantics.
`Stream.Close()` just calls `Stream.Dispose(true)` so as long as it's overloaded, `Close()` works as intended.

We could add a Stream extension that adds `Close()` wherever it's needed.

christophwille:
Having Close would reduce the amount of changes for existing consumers of the library. (and besides Close sounds a lot less destructive than Dispose... although everyone should be familiar with the pattern)

PR #264, opened 2018-08-20
Update README.md to reflect namespace change for ICSharpCode.SharpZipLib.Checksum

ronaldbarendse:
The checksum namespace changed from `ICSharpCode.SharpZipLib.Checksums.*` to `ICSharpCode.SharpZipLib.Checksum.*`, but the readme/documentation still referred to the plural name...

This should probably also be added as breaking change when upgrading from 0.86 to 1.0.0!


piksel:
Indeed. Thank you!

Issue #278, opened 2018-10-21
ZipStrings doesn't honor the CodePage property when it reads file entries


EvilBeaver:
Non-unicode filenames are extracted incorrectly because of behavior of ZipStrings.EncodingFromFlag function.

ZipStrings.EncodingFromFlag always uses SystemDefaultCodepage when it's not a Unicode. Here should be just a CodePage, which can be adjusted from client-side code.

I can fix this and send a Pull Request, but what is release cycle? Whether project abandoned? Maybe I'd better make a fork and use it?

Problem line:

https://github.com/icsharpcode/SharpZipLib/blob/master/src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs#L128

### Steps to reproduce
1. Open non-unicode archive with ciryllic symbols
2. Enumerate entries
3. See that they are incorrect

### Expected behavior
Symbols should be correct, when CodePage property is set to ciryllic one

### Actual behavior
Symbols are incorrect, changing CodePage property doesn't help

### Version of SharpZipLib
1.0

### Obtained from (place an x between the brackets for all that apply)
- Compiled from source
- Package installed using:
  - NuGet

piksel:
This is a duplicate of #274. The project is not abandoned, but I am the sole maintainer and have been very busy lately. A new version will be release pretty soon, so a PR would be welcomed.

piksel:
I am closing this due to the changes being merged.

Issue #274, opened 2018-09-21
Encoding can't be set as intended


Muffexx:
### Steps to reproduce
Encode ZipFile using 7Zip (including non ASCII characters like "ä", "ü")
Try to unzip using SharpZipLib with _CodePage_ set to "850".

### Expected behavior
Characters should be decoded correctly (including non ASCII characters).

### Actual behavior
Non ASCII characters aren't decoded correctly.

### Version of SharpZipLib
1.0.0

### Obtained from (place an x between the brackets for all that apply)
- Package installed using:
  - NuGet

### Comment
when looking at (in ZipStrings)

```.cs
private static Encoding EncodingFromFlag(int flags)
			=> ((flags & (int)GeneralBitFlags.UnicodeText) != 0)
				? Encoding.UTF8
                                : Encoding.GetEncoding(_SystemDefaultCodePage_)
```

I'm not sure whether this is as intended - since the _Codepage_ Property could be used instead of _SystemDefaultCodePage_.

The only way to get the characters decoded correctly is to encrypt via 7Zip with the cu-flag set to on (7-Zip uses UTF-8 for file names that contain non-ASCII symbols)
or setting the _SystemDefaultCodePage_ via reflection to a suitable codepage.

Maybe a setter for _SystemDefaultCodePage_ can be added
or _Codepage_ can be used in the mentioned code part?

Hope I didn't overlook anything.
regards

piksel:
Setting a custom encoding is not currently supported for extracting an archive, but it should be kind of trivial to add. Like adding a `CustomEncoding` property that, if set, is used instead of `SystemDefaultCodePage`.

A PR would be welcomed!

Also: I am a bit confused about your description, perhaps you could clarify this?

First in steps to reproduce:
> Encode ZipFile using 7Zip (including non ASCII characters like "ä", "ü")
> Try to unzip using SharpZipLib with CodePage set to "850".

But later on:
> (7-Zip uses UTF-8 for file names that contain non-ASCII symbols)

Why do you try to extract it with another encoding then?

Muffexx:
Yes, a custom encoding would be very helpful.

When using 7-Zip with default parameters it won't encode in UTF-8 by default (at least on my system).
Then when unzipping, UTF-8 is used by SharpZipLib since my local codepage won't support the special characters.
To resolve my issue I have three options:

- encode via 7-Zip and set codepage 850
- encode via 7-Zip and set UTF-8 to "on"
[...]

PR #280, opened 2018-10-25
Fix 274 and 278

EvilBeaver:

Issues #274 and #278 are fixed, but there's an another issue with Unicode test samples.

It seems, that Greek and Arabic samples are incorrect. I've added a workaround in https://github.com/EvilBeaver/SharpZipLib/commit/887e10def8eec0cd41fa531b9f4dd1f8c1c39227

Probably, old-code tests weren't failing, because ZipStrings.SystemDefaultCodePage was set to 65001, which is seems to be returned by default from Encoding.GetEncoding(0) on net.core

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Text;
 
 namespace ICSharpCode.SharpZipLib.Zip
@@ -29,8 +29,12 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// In practice, most zip apps use OEM or system encoding (typically cp437 on Windows). 
 		/// Let's be good citizens and default to UTF-8 http://utf8everywhere.org/
 		/// </remarks>
-		private static int codePage = Encoding.UTF8.CodePage;
+		private static int codePage = AutomaticCodePage;
 
+		/// Automatically select codepage while opening archive
+		/// see https://github.com/icsharpcode/SharpZipLib/pull/280#issuecomment-433608324
+		/// 
+		private const int AutomaticCodePage = -1;
 
 		/// <summary>
 		/// Encoding used for string conversion. Setting this to 65001 (UTF-8) will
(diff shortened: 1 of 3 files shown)


piksel:
This solution is more or less what I had intended too. The only problem with this approach is that you cannot force UTF-8 encoding by setting the Charset.

The SystemDefaultCodePage (`GetEncoding(0)`) differs on different frameworks and environments. I will take a look at the tests and see what is going on this weekend.

EvilBeaver:
> The only problem with this approach is that you cannot force UTF-8 encoding by setting the Charset.

No, forcing UTF-8 is still possible. If UTF-8 flag is set in archive - then UTF-8 will be used regardless of codepage property. Please, can you clear the problem you've mentioned? What utf-8 forcing do you want?

piksel:
With this solution you would not be able to unzip an archive that uses UTF-8 as it's encoding, whilst the unicode bit is not set. It would work if your systems default encoding is UTF-8, but not otherwise.
If we're adding support for setting the decompression encoding it would probably be better if we supported this scenario too. 

Perhaps having `Codepage` default to -1 which would indicate "Auto" (`SystemDefaultCodePage` or `UTF8` depending on Unicode bit), but if set to anything else it would use that codepage instead?

EvilBeaver:
> With this solution you would not be able to unzip an archive that uses UTF-8 as it's encoding, whilst the unicode bit is not set

Why? If client sets CodePage to 65001 then UTF-8 will be used and archive will be read correctly. Am I wrong?

piksel:
If CodePage is set to 65001 then
```.cs
CodePage == Encoding.UTF8.CodePage? SystemDefaultCodePage : CodePage
```
will result in `SystemDefaultCodePage` being used, no? And that is not UTF-8 in some environments.

EvilBeaver:
Yes, this is wrong. I'll fix it to -1 as you proposed before

EvilBeaver:
@piksel please review this.

EvilBeaver:
@piksel thanks a lot! When the new NuGet version is planned? I want to use this fix in my project as soon as possible.

piksel:
I'll try to get it done this weekend.

Issue #290, opened 2018-12-04
add or delete files error


a405012187:
### Steps to reproduce
1. i generate a compressed file
2. i want to update or delete some of them
3. 

### Expected behavior
files can successfully update or delete

### Actual behavior
1.some files success,they're all at the root dir,such as:"Test.txt"
2.some files error,they're all at subdirectory,such as : "Test\\Test.txt";and return "Cannot find entry to delete" exception.
3.when i scan the source code,i find that it maybe can't find the specified file name in the dictionary **updateIndex_** using the file name;but when i debug the program ,i find the file name is completly the same as the key;and i can also find the file using method **FindEntry** or **GetEntry**.

### Obtained from (only keep the relevant lines)
- Compiled from source, commit: _______
- Downloaded from GitHub √
- Package installed using NuGet

piksel:
Could you provide some code that describes your problem?

Version 0.86 is not supported, have you tried using v1.0?

a405012187:
> Could you provide some code that describes your problem?
> 
> Version 0.86 is not supported, have you tried using v1.0?

Here's  my code for adding file to the compressed file:
using (var zipStream = File.Open(zipPath, FileMode.OpenOrCreate, FileAccess.ReadWrite,
                        FileShare.ReadWrite))
                {
                    using (var zipFile = new ZipFile(zipStream))
                    {
                        zipFile.UseZip64 = UseZip64.On;
                        zipFile.IsStreamOwner = true;//the two properties have no effect
                        zipFile.BeginUpdate();
                        zipFile.Add(fileDataSource,entryName,CompressionMethod.Deflated);
                        //which inherited from the interface IStaticDataSource
                        zipFile.CommitUpdate();
                        zipFile.Close();                     
                    }
               }
when I add a file at the root directory,it works,such as "Test.txt";but when I update a file at a sub-directory which exist an old one,it shows two files with the same name,and the length of the compressed file is doubled,such as "Test\Test.txt".
ps:I've tried to installed the nuget package to my program ,but it shows that it can't be installed to a .NetFramework version=4.0 , and my program is only allowed to use the version 4.0

a405012187:
> Could you provide some code that describes your problem?
> 
> Version 0.86 is not supported, have you tried using v1.0?

btw,I can just find the DLL of 0.86 version,and v1.0 can only find a nuget-package which can not installed T^T 
Is v1.0 don't support .net 4.0?

Grunkles:
[...]

PR #295, opened 2018-12-18
Fix exception from deleting a directory ZipEntry in a ZipFile

benleedom:
`ZipFile.FindExistingUpdate(ZipEntry)` would always return -1 for directory `ZipEntry`s due to using `GetTransformedFileName` on all `ZipEntry`s. As a result, `ZipFile.Delete(ZipEntry)` would always throw an exception when passed a directory `ZipEntry`. The fix is to call `GetTransformedDirectoryName` for directory `ZipEntry`s.

I updated the test `AddAndDeleteEntries` in `ZipFileHandling.cs` to add and delete a directory; this test failed without the change to `ZipFile.FindExistingUpdate`.

I believe issue #290 is related.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -2461,7 +2461,9 @@ namespace ICSharpCode.SharpZipLib.Zip
 		private int FindExistingUpdate(ZipEntry entry)
 		{
 			int result = -1;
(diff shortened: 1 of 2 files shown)


piksel:
So, the consequence of this bug is that the entry name ends with a slash when created with `ZipFile.AddDirectory`, but not when it's searched for by `ZipFile.Delete` (which of course means it will not find it).

Test looks good. I'll try to release a bugfix update with this commit soon.

Issue #289, opened 2018-11-28
Performance deterioration in 1.0.0 in comparison to 0.86.0


decipherer:
We have been using version 0.86 of sharpziplib up till now. After updating to 1.0.0 we notice some welcomed improvements, but unfortunately also a big performance deterioration.

### Steps to reproduce
- Read a file using ZipInputStream in version 0.86 and record the duration
- Then read the same file using ZipInputStream in version 1.0 and record the duration
- Compare the two durations

### Expected behavior
The two durations should be roughly the same

### Actual behavior
The duration for version 1.0.0 is consistently a lot higher.
As an example, on my machine I notice the following difference in duration:

90MB file:
Version 0.86.0: 0.62s
Version 1.0.0: 3.3s

2GB file:
Version 0.86.0: 21.47s
Version 1.0.0: 81.31s

Both files were zipped using 7zip, using deflate, compression level normal , 32Kb Dictionary size, 32 Word size
Version 1.0.0 seems to be about 4-5 times slower in decompressing a file than version 0.86.0

### Version of SharpZipLib
0.86.0.518 vs 1.0.0.999

### Obtained from (only keep the relevant lines)
- Package installed using NuGet

piksel:
That's odd. Unfortunately it's hard to know exactly what has changed since 0.86 since there hadn't been a release for many years before 1.0.
Since you already have a good test case, would you mind running both cases through profiling?
I could not find any obvious changes that would yield such a drastic decrease in performance, although the switch to the new Stream API (IDisposable instead of Close) might be the culprit due to side effects somehow...

piksel:
Good catch!

Adler32 Hashing of 256 MiB using the current implementation: 
`7.5874 second(s)`

Using the optimized version that was present before the introduction of ArraySegment: 
`1.5796 second(s)`

I wrote a test case and a revert to the optimized version and will merge them before the next release!

decipherer:
Looking at the code it seems the Adler checksum is not used for (PK)Zip or GZip files. Therefore, it seems like unnecessary overhead to calculate the Adler checksum. Since the necessity of the Adler checksum can be easily deducted from the constructor arguments noZlibHeaderOrFooter (Deflater) and noHeader (Inflater) it would make sense not to calculate it when it's not necessary. Even for the more optimized version it can save quite some time.

piksel:
[...]

PR #301, opened 2019-01-08
Performance enhancement by skipping the (unnecessary) checks in ArraySegment

decipherer:
Another fix for this issue: https://github.com/icsharpcode/SharpZipLib/issues/289


piksel:
This is the same change as the one I made to Adler32, which is basically undoing the simplifications that the ArraySegment PR introduced. Since those introduce a significant slow down, the original (slightly more verbose) should be kept.

The reason for the failed checks is due to bad commit in master and not related to this PR.

Issue #144, opened 2016-10-06
ZipFile(Stream) constructor closes the stream before user can set IsStreamOwner


jaymrobinson:
### Steps to reproduce
1. Pass a stream containing a .pdf file to the ZipFile(Stream) constructor
2. Catch exception (ZipException: Cannot find central directory)
3. Attempt to access the stream again
### Expected behavior

The constructor shouldn't automatically close the stream on an exception.
At least, there should be an optional constructor parameter to specify IsStreamOwner.
### Actual behavior

The constructor closes the stream, so we can't do anything with it after the constructor fails.
### Version of SharpZipLib

0.86.0.518
### Obtained from (place an x between the brackets for all that apply)
- [ ] Compiled from source
  - branch: _______
  - commit: _______
- [ ] Downloaded DLL from GitHub
- [ ] Downloaded DLL from SourceForge
- [ ] Downloaded DLL from _______
- [ ] DLL included as part of
- Package installed using:
  - [ x] NuGet
  - [ ] MyGet
  - [ ] Chocolatey

Numpsy:
Hi,

I fell over this while doing some tests with the library, and thought i'd have a look at it.

Would something like https://github.com/Numpsy/SharpZipLib/commit/f1dddff99d5cdf97d0174847c5a425430ed5f29f be acceptable? (that avoids the issue in my use case).

I called the parameter 'leaveOpen' because that's what some other .NET classes (including the ZipArchive in System.IO.Compression) do.

Numpsy:
Fixed by #302 ?

Numpsy:
Think this one can be closed now?

piksel:
Indeed!

PR #302, opened 2019-01-13
Add a leaveOpen parameter to the ZipFile constructor, used to set the…

Numpsy:
… default value of isStreamOwner

Related to #144 - This change adds an extra version of the constructor for ZipFile which takes a 'leaveOpen' flag, which can be set to set the isStreamOwner flag during construction rather than after, which can help avoid unwanted stream disposal if the constructor throws (the param is called leaveOpen rather than isStreamOwner to match the naming of various other .net libs, and could be changed).

[WIP] because it could do with some tests if deemed a reasonable thing to do.


piksel:
Good choices! Some simple unit tests for new constructors should probably be added.

Numpsy:
added a few tests

Issue #303, opened 2019-01-14
Blob stored in Azure with AES password protected


eldridgemartinez:
### Steps to reproduce
1. Create a zip in Winrar WITH a password. I'm using Winrar so that the default password in AES instead of the ZipLegacy. Make sure one of the files inside the zip is at least 17KB. 
2. Upload the file to Azure as BlockBlob
3. Create a sample program that uses 'Microsoft.WindowsAzure.Storage' SDK and calls the OpenRead() method passed to the new ZipFile() constructor.  
  Important Note: Set the StreamMinimumReadSizeInBytes property of CloudBlockBlob object to 16384 (or any value less than one of the files inside your zip)

An example code (also in [ZipTest.zip](https://github.com/icsharpcode/SharpZipLib/files/2754376/ZipTest.zip) the complete project):
```cs
using ICSharpCode.SharpZipLib.Core;
using ICSharpCode.SharpZipLib.Zip;
using Microsoft.WindowsAzure.Storage;
using System;
using System.IO;

namespace ZipTest
{
    class Program
    {
        static void Main(string[] args)
        {
            var blobConString = @"<replace-with-your-azure-conn-string>";
            var storageAccount = CloudStorageAccount.Parse(blobConString);
            var myClient = storageAccount.CreateCloudBlobClient();
            var container = myClient.GetContainerReference("test");
            var blockBlob = container.GetBlockBlobReference("TestZip-pass-is-testing.zip");

            // Mininum is 16KB.. This needs to be smaller to one of the files in the
            // zip in order to replicate the issue
            blockBlob.StreamMinimumReadSizeInBytes = 16384;

            var zipFile = new ZipFile(blockBlob.OpenRead())
            {
                Password = "testing"
            };

            var outputFolder = $@"c:\temp\out\{Guid.NewGuid()}";
            WriteToDisk(zipFile, outputFolder);

            Console.WriteLine($"Zip file extracted to {outputFolder}. Press any key to exit.");
            Console.ReadKey();
        }

        private static void WriteToDisk(ZipFile zipFile, string outFolder)
        {
            foreach (ZipEntry zipEntry in zipFile)
            {
                if (!zipEntry.IsFile)
                    continue;

                var fileName = Path.GetFileName(zipEntry.Name);
                var fullZipToPath = Path.Combine(outFolder, fileName);

                var directoryName = Path.GetDirectoryName(fullZipToPath);
                Directory.CreateDirectory(directoryName);

[...]

Issue #181, opened 2017-06-09
256 bit AES Encryption doesn't work


mattosaurus:
256 bit AES encryption doesn't work, an error message of "System.ArgumentException: 'Specified initialization vector (IV) does not match the block size for this algorithm.'" is generated on line 87 of ZipAESTransform.cs during the creation of the encryptor. 128 bit AES encryption works as expected.

```C#
_encryptor = rm.CreateEncryptor(byteKey1, byteKey2);
```

### Steps to reproduce
Try and encrypt file with 256bit AES encryption

```C#
class Program
{
    static void Main(string[] args)
    {
        ZipFile(@"C:\TEMP\test\test.txt", @"C:\TEMP\test\test.zip", "password");
    }

    public static void ZipFile(string srcPath, string outPath, string password)
    {
        using (ZipOutputStream zipStream = new ZipOutputStream(File.Create(outPath)))
        {
            zipStream.SetLevel(9); // 0-9, 9 being the highest level of compression

            // optional. Null is the same as not setting. Required if using AES.
            zipStream.Password = password; 

            ZipEntry zipEntry = new ZipEntry(Path.GetFileName(srcPath));

            // Specifying the AESKeySize triggers AES encryption. 
            // Allowable values are 0 (off), 128 or 256. 
            // A password on the ZipOutputStream is required if using AES.
            zipEntry.AESKeySize = 256; 
            zipEntry.DateTime = DateTime.Now;

            zipStream.PutNextEntry(zipEntry);

            byte[] buffer = new byte[4096]; // Zip the file in buffered chunks

            using (FileStream fileStream = File.OpenRead(sourceFilePath))
            {
                // Using a fixed size buffer here makes no noticeable difference
                // for output but keeps a lid on memory usage.
                int sourceBytes = fileStream.Read(buffer, 0, buffer.Length);

                while (sourceBytes > 0)
                {
                    zipStream.Write(buffer, 0, sourceBytes);
                    sourceBytes = fileStream.Read(buffer, 0, buffer.Length);
                }
            }
        }
    }
}
```

### Expected behavior
Password protected zip file should be created.

### Actual behavior
Fails to encrypt zip file.

### Version of SharpZipLib
0.86.0.518

### Obtained from (place an x between the brackets for all that apply)
- Compiled from source
 - branch: master
 - commit: be53259

fabien367:
Hello, for me it's the same.
[...]

PR #304, opened 2019-01-14
Read more bytes if returned count does not match the requested length…

eldridgemartinez:
…. End of file also returns 0 so its non-breaking

### DIFF
--- a/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
+++ b/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -10,9 +10,9 @@
   
   <!-- Nuget specific tags -->
   <PropertyGroup>
-    <AssemblyVersion>1.1.0.6</AssemblyVersion>
-    <FileVersion>1.1.0.6</FileVersion>
-    <Version>1.1.0</Version>
+    <AssemblyVersion>1.1.1.0</AssemblyVersion>
+    <FileVersion>1.1.1.0</FileVersion>
+    <Version>1.1.1</Version>
     <PackageId>SharpZipLib</PackageId>
     <Company>ICSharpCode</Company>
     <Authors>ICSharpCode</Authors>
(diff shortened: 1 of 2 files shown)


eldridgemartinez:
Please refer https://github.com/icsharpcode/SharpZipLib/issues/181 for details of this PR.

piksel:
Don't you mean #303? #181 has nothing to do with the amount of bytes read from the input stream...

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -4038,11 +4038,27 @@ public override int Read(byte[] buffer, int offset, int count)
 						baseStream_.Seek(readPos_, SeekOrigin.Begin);
 					}
 					int readCount = baseStream_.Read(buffer, offset, count);
+
+					// Azure CloudBlockBlob SDK Stream implementation is: Returns only the bytes in its buffer and does not fetch from remote until its buffer is empty.
+					// This observation causes an issue in above baseStream_.Read call because the buffer is only filled partially
+					int correctReadCount = readCount;
```
piksel: This file works according to spec. If the underlying stream returns fewer bytes, it correctly returns the same amount of bytes. The changes does not belong here (and neither does the comment about third-party specific details).

src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
```diff
@@ -10,9 +10,9 @@
   
   <!-- Nuget specific tags -->
   <PropertyGroup>
-    <AssemblyVersion>1.1.0.6</AssemblyVersion>
```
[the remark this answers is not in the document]
eldridgemartinez: Oh sorry. I was doing my test build and forgot to cleanup commits. I will resolve this. Thanks

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
eldridgemartinez: I agree that the change can be applied in the caller ZipAESStream.cs file too. Thanks for looking and let me know if you applied that fix and I will help test. You can ignore this PR though.

eldridgemartinez:
Hi, thanks for looking and pointing to where should the fix go. I agree with the fix to be applied in ZipAESStream.cs to read more and dont treat the returned lesser bytes as end of file. I will also try to see ans fix there. I badly needed this fix as we are closing to production release. Thank you

eldridgemartinez:
hi, I close this PR and create a fresh one. I'll send in a while after I'm done testing.. I follow your suggestion to fix on ZipAESStream.cs.

PR #305, opened 2019-01-15
read more bytes to fill the required length

eldridgemartinez:
Please refer to issue #303  for complete replication

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Encryption/ZipAESStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Encryption/ZipAESStream.cs
@@ -81,8 +81,16 @@ namespace ICSharpCode.SharpZipLib.Encryption
 				int obtained = _stream.Read(_slideBuffer, _slideBufFreePos, lengthToRead);
 				_slideBufFreePos += obtained;
 
-				// Recalculate how much data we now have
-				byteCount = _slideBufFreePos - _slideBufStartPos;
+                if (obtained > 0 && obtained < lengthToRead)
+                {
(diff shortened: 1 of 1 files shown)


src/ICSharpCode.SharpZipLib/Encryption/ZipAESStream.cs
```diff
@@ -81,8 +81,16 @@ public override int Read(byte[] buffer, int offset, int count)
 				int obtained = _stream.Read(_slideBuffer, _slideBufFreePos, lengthToRead);
 				_slideBufFreePos += obtained;
 
-				// Recalculate how much data we now have
-				byteCount = _slideBufFreePos - _slideBufStartPos;
+                if (obtained > 0 && obtained < lengthToRead)
+                {
+                    // less bytes are returned but may not be the end of file. Get more to fill the required length. Eof returns 0 bytes so this is safe
+                    int moreObtained = _stream.Read(_slideBuffer, _slideBufFreePos, lengthToRead - obtained);
```
MarcWils: Why only 1 extra read? You could try reading from the inner stream until the buffer is completely filled or the inner stream returns 0 bytes. Seems to be more solid.

eldridgemartinez: It needs one extra read to fill the required length else it throws error and I don't want to change the logic of lengthToRead. This one works and I tested it. Can you help me how to fix this?

eldridgemartinez: Yes thats good. I will cancel my PR if that is working and merged. Let me know if there is a new nuget release. Thank you

PR #309, opened 2019-01-21
 Fix ZipEntry name mismatch when attempting to update a directory entry

dotlive:

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -1579,7 +1579,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		{
 			contentsEdited_ = true;
 
-			int index = FindExistingUpdate(update.Entry.Name);
+			int index = FindExistingUpdate(update.Entry);
 
 			if (index >= 0)
 			{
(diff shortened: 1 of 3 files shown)


src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -1579,7 +1579,7 @@ private void AddUpdate(ZipUpdate update)
 		{
 			contentsEdited_ = true;
 
-			int index = FindExistingUpdate(update.Entry.Name);
+			int index = FindExistingUpdate(update.Entry);
```
piksel: Why is FindExistingUpdate changed here?

The `FindExistingUpdate(ZipEntry)` method should probably just call `FindExistingUpdate(string)` for DRYness, but that's for another issue...

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -2511,6 +2513,15 @@ private int FindExistingUpdate(string fileName)
 			return result;
 		}
 
+		private bool IsDirectoryName(string name)
```
piksel: `GetTransformedFileName` is only referenced in `FindExistingUpdate`, perhaps `GetTransformedFileName`, `GetTransformedDirectoryName` and `IsDirectoryName` should be reduced to a single function called `GetTransformedEntryName`? The two former ones are almost identical.

piksel:
Closing due to lack of response.

PR #312, opened 2019-01-24
Change value of CompressionMethod.Bzip2 from 11 to 12

Numpsy:
CompressionMethod.BZip is curently defined as 11, but it seems that it should actually be 12


piksel:
As per APPPNOTE.TXT:4.4.5, 11 is reserved and 12 is BZip2

PR #313, opened 2019-01-25
When invoking 7-zip in the tests, put quotes around the path to the test file

Numpsy:
When running the unit tests, I've seen a couple of failures in the encryption tests that seem to be simply because the temp file path has a space in it, which causes issues when invoking 7-zip on the file.

Perhaps it needs quotes putting around the test file path?


piksel:
I designed that test to work in the CI environment and did not consider the case where the user have a space in their name.
If this solves that case then I'd be happy to merge it.

Numpsy:
Yes, that was what caused it (just happened to run the tests on a new system with a space in the user name, and 2 of the AES tests failed even though 7-zip was installed)

PR #314, opened 2019-01-27
Fix writing extra data when updating Zip64 entries

### LINKED ISSUE #310: Zip64 cannot be delete when it has more than one file
### Steps to reproduce
1.Create a zip64 which has multi file
2.BeginUpdate()
3.Delete one of the file ( Delete(ZipEntry entry) )
4.CommitUpdate()

### Expected behavior
Without Error

### Actual behavior
ZipException: Extra data extended Zip64 information length is invalid

### Version of SharpZipLib
Ver. 1.1.0

Numpsy:
WriteCentralDirectoryHeader currently checks IsZip64Forced when deciding to write -1 to the base size fields [ZipFile.cs#L2174](https://github.com/Numpsy/SharpZipLib/blob/0fff917d8ef61e9cd0b9b8574301498a6416e0bf/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L2174) but not when writing the extra data [ZipFile.cs#L2208](https://github.com/icsharpcode/SharpZipLib/blob/4f541a404f324bd882bc1e8ecd1182c05909b08a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L2208), so the length doesn't get written.

This changes it to check in both places, and adds a test

Assuming that the length field and extra data should always be in sync (would you ever set size to -1 and not write the extra data?), might be better to set some 'wantsExtraSize' flags rather than checking ((entry.Size >= 0xffffffff) || otherFlags) in two places?

Closes #310.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -2171,8 +2171,10 @@ namespace ICSharpCode.SharpZipLib.Zip
 				WriteLEInt((int)entry.Crc);
 			}
 
+			bool useExtraCompressedSize = false; //Do we want to store the compressed size in the extra data?
 			if ((entry.IsZip64Forced()) || (entry.CompressedSize >= 0xffffffff))
 			{
+				useExtraCompressedSize = true;
 				WriteLEInt(-1);
 			}
 			else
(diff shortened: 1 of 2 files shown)


piksel:
> would you ever set size to -1 and not write the extra data?
No. At least not if you want to comply with the spec ([APPNOTE](https://www.pkware.com/documents/casestudies/APPNOTE.TXT) 4.4.1.4).
It is just a recommendation to set the size to -1, but in doing so you opt in to also add the Zip64 extended length (as I interpret it).
So yeah, adding a variable is indeed better (although I prefer not to anthropomorphize code :), maybe `useExtraSize`).
It could of course be argued that if you want to force the zip64 field (for compliance with some inflater that only uses that field or whatever), and the size is < 65535, the non-extended size should still be the actual value.

piksel:
Ah, I missed that. Looks good!

Issue #322, opened 2019-02-18
System.ArgumentNullException thrown from ZipOutputStream when writing AES encrypted/Stored entries


Numpsy:
### Steps to reproduce
Try to use ZipOutputStream to create an AES encrypted entry whose compressionMethod is set to Stored.
e.g.
```
using (var memoryStream = new MemoryStream())
{
	using (var zipStream = new ZipOutputStream(memoryStream))
	{
		zipStream.IsStreamOwner = false;
		zipStream.Password = "password";  // optional. Null is the same as not setting. Required if using AES.

		ZipEntry zipEntry = new ZipEntry("stored");
		zipEntry.AESKeySize = 256;
		zipEntry.DateTime = DateTime.Now;
		zipEntry.CompressionMethod = CompressionMethod.Stored;

		zipStream.PutNextEntry(zipEntry);

		byte[] dummyData = Encoding.UTF8.GetBytes(DummyDataString);

		using (var dummyStream = new MemoryStream(dummyData))
		{
			dummyStream.CopyTo(zipStream);
		}

		zipStream.CloseEntry();
	}
}	
```

### Expected behavior
The entry should be added

### Actual behavior
The call to CloseEntry() throws an ArgumentNullException.

### Obtained from (only keep the relevant lines)
- Compiled from source, commit: 25ee644

Numpsy:
Problem seems to be that the AESAuthCode property of the output stream isn't set up in this case, so the attempt to use it @ https://github.com/icsharpcode/SharpZipLib/blob/25ee6440f444aa326ee8f0b98e6223296f35fa6d/src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs#L493 throws.

See referenced PR for a possible fix.

Numpsy:
Should be fixed by #323, so i'll close this now.

PR #323, opened 2019-02-18
Change ZipOutputStream.CloseEntry() to work for AES encrypted Stored entries

Numpsy:
Possible fix for #322 - Change ZipOutputStream.CloseEntry() to set up AESAuthCode itself when the compression method is 'Stored' (it's normally done by DeflatorOutputStream.Finish(), which isn't called in this case).


piksel:
This seems like an OK solution. The authcode should probably not be handled by Deflate in the first place, but rather in another layer. But that would be a much larger restructure.

PR #325, opened 2019-02-19
Ability to mimic a zip file created by a Linux file-system.

bjorg:
# Use Case
Create a zip archive with specific permissions, such as the executable flag, using the `ExternalAttributes` property. The Linux `unzip` command only applies such permissions if the zip entry indicates the file was added by a Linux host-system.

# Solution
Expose the more specific `ZipEntry` constructors that allow controlling the version number and host-system information. Ensure this information is used when creating the zip archive.

# Changes:
1. Changed access level from `internal` to `public` on constructors for `ZipEntry`. This allows setting the `madeByInfo` value.
1. Changed `VersionMadeBy` property to return the entire `ushort`. This is required by `ZipFile` and `ZipOutputStream` to write the correct value.
1. Fixed `HostSystem` property to clear out the upper byte where the host system information is stored.
1. Changed `ZipFile` and `ZipOutputStream` to respect the `VersionMadeBy` value of the `ZipEntry` instance instead of hard-coding it.

# Agreement 

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
@@ -169,7 +169,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// <exception cref="ArgumentNullException">
 		/// The name passed is null
 		/// </exception>
-		internal ZipEntry(string name, int versionRequiredToExtract)
+		public ZipEntry(string name, int versionRequiredToExtract)
 			: this(name, versionRequiredToExtract, ZipConstants.VersionMadeBy,
 			CompressionMethod.Deflated)
 		{
@@ -192,7 +192,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// This constructor is used by the ZipFile class when reading from the central header
 		/// It is not generally useful, use the constructor specifying the name only.
 		/// </remarks>
-		internal ZipEntry(string name, int versionRequiredToExtract, int madeByInfo,
+		public ZipEntry(string name, int versionRequiredToExtract, int madeByInfo,
 			CompressionMethod method)
 		{
 			if (name == null)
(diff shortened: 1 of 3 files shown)


piksel:
I would prefer to add this by having a "host OS override" that can be set and which in turn sets the top version byte.

bjorg:
There is a `HostSystem` property that might fit the bill. It had a bug where it cleared out the high-byte instead of the low-byte ([see diff](https://github.com/icsharpcode/SharpZipLib/pull/325/files#diff-5793f364c4e98ef0037a63c6b250ca08R520)).

The big problem, IMO, is the `VersionMadeBy` property. As implemented, it always truncates the highg-bytes ([see diff](https://github.com/icsharpcode/SharpZipLib/pull/325/files#diff-5793f364c4e98ef0037a63c6b250ca08L447)). Fixing that could potentially be a breaking change since dependent code may not expect the high-byte to have a value.

If `VersionMadeBy` cannot be fixed, then another property needs to be added that holds both the OS and zip version information. This information is later needed by `ZipFile.WriteCentralDirectoryHeader(...)` ([see diff](https://github.com/icsharpcode/SharpZipLib/pull/325/files#diff-90f6fd221702a6a05926e8890e900c78R2160)) and by `ZipOutputStream.Finish(...)` ([see diff](https://github.com/icsharpcode/SharpZipLib/pull/325/files#diff-d8584e0c743bf2ed9d16a9b9f73c46e8R747)).

Let me know how you would like me to proceed and I'll update the pull-request. Thanks!

piksel:
So, wouldn't just fixing the HostSystem property be enough then? 
And using HostSystem + VersionMadeBy in ZipFile and ZipOutputStream?

bjorg:
I think so, yes. That would probably work. Let me give it a shot and update the pull-request.

bjorg:
@piksel 
I updated the code. It now just fixes the `HostSystem` property bug and then ORs it with `VersionMadeBy` property in-lieu of the `ZipConstants.VersionMadeBy` constant.

piksel:
As it turns out, it's not possible to set the `HostSystem` property of an entry using `ZipFile` (unless the file is empty).
I made some tests and added this functionality in a new PR.

Issue #27, opened 2013-12-29
Tar ASCII translation does not work


dgrunwald:
[SD-744](http://bugtracker.sharpdevelop.net/issue/ViewIssue.aspx?id=744), originally created on 4/2/2006 22:23:38 by John Reilly

I have met just a little prbl whith TAR.GZ files when I set the
asciiTranslate option on "true". When this option is set on false,
everything works great, but when it is on true, an IOException appears
:(

Here is the full error message :
__________________________________
Une exception non gérée du type 'System.IO.IOException' s'est produite
dans mscorlib.dll

Informations supplémentaires : Le processus ne peut pas accéder au
fichier "C:\Documents and Settings\Jeje\Mes documents\Visual Studio
Projects\TestXLAccess\bin\Debug\Test\batchs\purge.sh", car il est
en cours d'utilisation par un autre processus.
_______________________________________

I guess you prefer in English ;-)
__________________________________
A non handeled exception of  'System.IO.IOException' type has occured in
mscorlib.dll

Further information : The process cannot access to the file
"C:\Documents and Settings\Jeje\Mes documents\Visual Studio
Projects\TestXLAccess\bin\Debug\Test\batchs\purge.sh", because it
is being used by an other process
_______________________________________

So, I think SharpZipLib might create a new process to write into text
files but there is a colision with the main process which has created
the file or something like that.

If you have any idea about how this little bug could be resolved, I
would be glad to hear it :)

Thx again for your work.
Jerome

piksel:
This should have been fixed by #326

Numpsy:
was the translate option on extract actually fixed? I had a recollection that there was still an issue with it trying to read a file that it had locked for write, though I might be misremembering.

piksel:
Well, both yes and no :3

It *was* fixed as it is no longer trying to open the file twice, so no exception.
But it's still checking if the *destionation* file is ASCII, which will never be true, so that part is still wrong (which simply means that we don't do any translation).

PR #326, opened 2019-02-21
Change TarArchive.ExtractContent() to dispose of output streams in case of an exception

Numpsy:
refs #321 - change TarArchive.ExtractEntry to dispose of output file streams in a finally block, in case an exception occurs. 
(pretty simple change, but it looks like changing the indentation makes the diff viewer make the change look greater than it is).

Also adds a unit test that failed previously and passes now.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs
+++ b/src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs
@@ -623,20 +623,32 @@ namespace ICSharpCode.SharpZipLib.Tar
 
 				if (process)
 				{
-					bool asciiTrans = false;
-
-					Stream outputStream = File.Create(destFile);
-					if (this.asciiTranslate)
+					using (var outputStream = File.Create(destFile))
 					{
-						asciiTrans = !IsBinary(destFile);
+						if (this.asciiTranslate)
+						{
+							// May need to translate the file.
+							ExtractAndTranslateEntry(destFile, outputStream);
+						}
+						else
+						{
+							// If translation is disabled, just copy the entry across directly.
+							tarIn.CopyEntryContents(outputStream);
+						}
 					}
+				}
+			}
+		}
 
-					StreamWriter outw = null;
-					if (asciiTrans)
-					{
-						outw = new StreamWriter(outputStream);
-					}
+		// Extract a TAR entry, and perform an ASCII translation if required.
+		private void ExtractAndTranslateEntry(string destFile, Stream outputStream)
+		{
+			bool asciiTrans = !IsBinary(destFile);
 
+			if (asciiTrans)
+			{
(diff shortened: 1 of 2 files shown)


Numpsy:
I think at the time I was just trying to make the minimal change to fix the issue.

Possibly the using approach has issues with the way the StreamWriter is only optionally created, but when it is it takes ownership of the output stream such that you don't have to dispose both?
Would have to have another look, perhaps it could do with being refactored to improve that.

piksel:
StreamWriter has a `leaveOpen` parameter, but you have to specify all four parameters:
```public StreamWriter (System.IO.Stream stream, System.Text.Encoding encoding, int bufferSize, bool leaveOpen);```  
Even if the change is small, it's hard to tell what has changed due to the scope change (as you said).

Numpsy:
Hmm, I didn't pay too much attention to what this function was actually doing when I looked at it before, but on a closer look it seems a bit wrong in general. I mean, 

1) The IsBinary check @ https://github.com/icsharpcode/SharpZipLib/blob/27ac139f213a8a2f723b43418130b0098b7e4080/src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs#L631 looks like it's trying to check if the output file is binary, rather than the input data?

2) What's the loop @ https://github.com/icsharpcode/SharpZipLib/blob/27ac139f213a8a2f723b43418130b0098b7e4080/src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs#L653 going to do if there aren't any new lines in the buffer it's going through?

piksel:
I think what it's trying to do is mimic `fopen()` without the `b` mode flag. Basically translating line endings to the target operating systems preferred ones.  
They probably copied the `isBinary` from the archive creation code where it makes much more sense. This test should always return false since it tries to find "binary" data in truncated file. Either way it's wrong.  
  
The loop would just not write anything to the output  if no newlines are encountered. Which also means that the last line would always be omitted.  
Instead of this code, simply using `StreamReader.ReadLine()` followed by `StreamWriter.WriteLine()`  would replace the line endings.

Numpsy:
that seems more reasonable (was also thinking that the existing code will have issues if the input already has Windows line endings, and that would fix that as well).

This is going on a bit of a tangent though - i'll refactor the stream disposal to use usings, and then see about a seperate issue for the rest.

Numpsy:
I've changed it to use using blocks and also to just use TarInputStream.CopyEntryContents in the non-translate cases.
I haven't made any changes to the binary detection or line handling in this set - will open a seprate issue about that.

Numpsy:
Actually, I wonder if the usage of IsBinary there is actually the cause of #27, where trying to open the file for read while it's already open for create/write caues the exception?

piksel:
Good catch! That actually also answers the question "how did this ever work?". It's probably never even turned on.  
Since the write file handle is always acquired (without Read sharing permitted) this would always fail (on windows).

PR #349, opened 2019-06-13
[Feature] Add support at .net 3.5  Framework target 

workgroupengineering:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,7 @@ dotnet_csproj:
   informational_version: $(VERSION)
 install:
 - ps: tools/appveyor-install.ps1
+- ps: tools/appveyor-vs-add-component
 nuget:
   project_feed: true
   disable_publish_on_pr: true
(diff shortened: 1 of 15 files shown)


christophwille:
What is the point of FW 3.5 support in 2019? Making our code base and build processes more complicated for absolute edge cases is not helpful (we just removed 3.5 support from AvalonEdit for good).

workgroupengineering:
Hi @christophwille 
sorry if I didn't open a problem before creating a pr.

> What is the point of FW 3.5 support in 2019? Making our code base and build processes more complicated is not helpful (we just removed 3.5 support from AvalonEdit for good).

FW 3.5 at now is more diffuse version of .NET Framework. It' is running from oldest win xp at new windows 10.
There is a lot of software compiled with FW 3.5 that cannot be updated since proprietary libraries are not available to the new version of framework. Especially in machine automation.

christophwille:
Yes, it is old. And that is precisely my point - it makes modernizing the library hard. No, we won't re-add support for this ancient framework.

workgroupengineering:
I respect your decision and I will close the PR, but I am not of your opinion.

If an old man continues to do his job well, do I have to treat him if he has a problem?

piksel:
Well, the problem here is that it doesn't do the job equally well. This PR is in itself testament to that. You need workarounds, polyfills and preprocessor directives to get it to work. This means that we have to support code that we normally never run, have to do additional polyfills / backports for new features and as a result, spend less time doing actual development.

PR #351, opened 2019-06-15
Add support for Filename field in GZip

piksel:


src/ICSharpCode.SharpZipLib/GZip/GzipInputStream.cs
```diff
@@ -169,149 +181,108 @@ private bool ReadHeader()
...
+					{
+						fname[fnamePos++] = (byte)readByte;
+					}
 					headCRC.Update(readByte);
 				}
 
-				if (readByte < 0)
-				{
-					throw new EndOfStreamException("EOS reading GZIP header");
-				}
```
piksel, line 292: `ReadLeByte()` never returns anything but a byte, it will itself throw an exception in the case of an EOS

Issue #341, opened 2019-05-07
Invalid Characters in Zip throws exception


RevenantBob:
### Steps to reproduce
1. Use a Zip file produced on a *nix system or MacOS (In this case)
2. Zip files with invalid windows characters that are valid on MacOS/Linux
3. Open up Zip archive with SharpZipLib API

Exception is thrown.

### Expected behavior
SharpZipLib should not load a ZipFile class and try to clean up the entry filenames before returning them to the user. The user should be responsible for cleaning up entries themselves in whatever method they see fit.

### Actual behavior
SharpZipLib throws an exception in ZipEntry line 1303 because IsPathRooted checks for invalid characters. It also removes root path names.

### Version of SharpZipLib
1.10

### Obtained from (only keep the relevant lines)
- Package installed using NuGet

For now I worked around the issues by removing the CleanName function from ZipEntry in my own local copy.

piksel:
Version 1.10? I'm guessing it's 1.1.0?

Numpsy:
Confirmed with the current source using a zip file which contains a file that has question marks in the name - it throws a System.ArgumentException from Path.IsPathRooted.

Given that the comments on the CleanName function talk about converting names into zip formats, doesn't sound like something that should be called on extraction?
The ZipEntry constructor comment at https://github.com/icsharpcode/SharpZipLib/blob/f03a2ef068977ebfaad376335a2db524278d8813/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs#L163 also says that the name format isn't enforced, so it might be misleading that the constructor chain does then call CleanName?

piksel:
Agreed. It should at least be platform-specific, but then again, we could just let the file system throw the exception on invalid file names instead.

Numpsy:
There's a bunch of functionality in the NameTransform bits for processing invalid characters and such that the caller and/or container could use rather than ZipEntry doing it, including some alternate path root handling code that CleanName could possibly use instead of IsPathRooted if required?

Related: Most of the Add() functions on ZipFile create ZipEntrys using the EntryFactory, which runs the names through NameTransform, but the variants that actually take ZipEntries don't.
Would removing the CleanName calls from the ZipEntry constructor then allow you to add entries with invalid names to ZipFile, in situations where you perhaps couldn't before? (or is that just up to the caller to deal with?)

piksel:
[...]

PR #362, opened 2019-06-22
Don't call CleanName from the ZipEntry constructor.

Numpsy:
Possible fix for #341 - don't call CleanName in the ZipEntry constructor.

This matches the doc comments on one of the ZipEntry constructors which says that the name format isn't enforced anyway. Not sure if this causes any gaps where names won't be validated on creating zips - not sure whether it'd be better to do some validation for 'new' ZipEntries, but just not for ones created internally when reading an existing zip?

Also marked as WIP because on further testing it seems that a lot of the validation for the characters in paths has been removed from .NET Core, meaning that the exception described in #341 might only happen in the full version of .NET (and as the unit tests are only built as Core, the exception won't be seen there).

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
@@ -211,7 +211,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 			}
 
 			this.DateTime = DateTime.Now;
-			this.name = CleanName(name);
+			this.name = name;
 			this.versionMadeBy = (ushort)madeByInfo;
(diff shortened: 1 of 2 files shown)


hoborm:
Hi!
Do you have any roadmap for the next release including this MR or do you suggest to build and package it to a NuGet package if I'd like to use it? I don't want to hurry the release, I just would like to know the plans.

Numpsy:
I don't know what the plans for the next release are, but you could maybe look at the CI builds if you need a build of it?

hoborm:
> 
> 
> I don't know what the plans for the next release are, but you could maybe look at the CI builds if you need a build of it?

Oh I didn't notice that it's already available in pr. Thank you once again! :)

Issue #357, opened 2019-06-18
Zip64 Central Directory skipped if not explicitly needed


piksel:
Other libraries handle Zip64 differently than SharpZipLib. This leads to issues when the library doesn't react the same way to non-compliant archives.  
  
The library should instead always attempt to find a Zip64 signature.

Numpsy:
Related to #291 as well?

Numpsy:
Should be fixed by #363 ?

piksel:
Yes

Issue #356, opened 2019-06-18
ICSharpCode.SharpZipLib.Zip.ZipException: Wrong Central Directory signature


DmitriyKirakosyan:
File to test: https://dl.google.com/android/repository/sys-img/android-wear/x86-P_r02.zip.
SharpZipLib fails to unzip this file, while Windows10 is able to do it using `Extract All...` menu item.

### Steps to reproduce
1. Download the test file;
2. Try to unzip it using SharpZipLib;
3. Observe the `Wrong Central Directory signature` error.

### Version of SharpZipLib
1.1.0

### Obtained from (only keep the relevant lines)
- Package installed using NuGet

piksel:
Confirmed. This is due to the fact that the file has a Zip64 Central Directory, while not using any Zip64 extensions. It also contains a non-zip64 CD, but it seems to be corrupt.  
  
The way SharpZipLib detects whether to use Zip64 is if any of the fields are set to all 1s (0xffff etc), and only then tries to find the Zip64 signature.  
This means that  `unzip` and other programs that just tries to find both headers at once works fine, while SharpZipLib (albeit correctly) complains about the invalid Central Directory.  
  
We should probably attempt to find the Zip64 signature at once as well, this would my the library work more like most other zip-implementations and handle non-compliant zip files better.

piksel:
I made a new issue #357 for the root cause to link this and other issues into.

PR #363, opened 2019-06-23
Change ZipFile.ReadEntries to always look for the Zip64 central directory

Numpsy:
Possible change for #357 - in ZipFile.ReadEntries, always look for the zip64 central directory and use it if present, rather than only looking if we think we need it.
This change will still throw if it thinks the zip64 directory is required but can't find it (same logic there as before).

The change is sufficient to get the file from #356 loading successfully without causing any issues with the existing unit tests, but doesn't add any more tests as I don't have an appropriate test file to hand (the one linked from #356 being too large to use for a unit test)


piksel:
I know how to manually create a corrupt archive, so I will be adding a test

piksel:
Finally got the test working :D

PR #364, opened 2019-06-25
add encoding parameter to creating tar entry

itn3000:
This PR is similar to #182, but this PR adds System.Text.Encoding to TarEntry related API.
default is same as current master behavior(omit upper byte) because more faster than using encoding if tar archive assured having ASCII filename only.


itn3000:
Codacy says "`TarInputStream.CreateEntry(byte[] headerBuffer, Encoding enc)` should be static method".
But other `TarInputStream.CreateEntry` functions are not static, should I `CreateEntry(byte[] headerBuffer, Encoding enc)` make static?

piksel:
The problem right now with Codacy is that the current code right now is way beyond bad.  
I think it's still better to be consistent here than to adhere to Codacy since it's better to do a full cleanup in another commit of all the functions.

piksel:
As stated in #182:
The specification suggests using extended headers for encodings other than ASCII, but GNU Tar writes the UTF8-encoded bytes as the Name so I guess we should too.

This should be fine.

src/ICSharpCode.SharpZipLib/Tar/TarOutputStream.cs
```diff
@@ -252,23 +287,23 @@ public void PutNextEntry(TarEntry entry)
...
 				buffer.WriteBlock(blockBuffer);  // Add special long filename header block
 
 				int nameCharIndex = 0;
 
-				while (nameCharIndex < entry.TarHeader.Name.Length + 1 /* we've allocated one for the null char, now we must make sure it gets written out */)
+				while (nameCharIndex < namelen + 1 /* we've allocated one for the null char, now we must make sure it gets written out */)
 				{
 					Array.Clear(blockBuffer, 0, blockBuffer.Length);
-					TarHeader.GetAsciiBytes(entry.TarHeader.Name, nameCharIndex, this.blockBuffer, 0, TarBuffer.BlockSize); // This func handles OK the extra char out of string length
+					TarHeader.GetAsciiBytes(entry.TarHeader.Name, nameCharIndex, this.blockBuffer, 0, TarBuffer.BlockSize, encoding); // This func handles OK the extra char out of string length
```
piksel, line 300: This should probably be renamed?

src/ICSharpCode.SharpZipLib/Tar/TarOutputStream.cs
```diff
@@ -475,6 +510,11 @@ private void WriteEofBlock()
 		/// </summary>
 		protected Stream outputStream;
 
+		/// <summary>
+		/// name encoding
```
piksel, line 514:
```suggestion
		/// The <see cref="Encoding"/> used for the Name fields
```

test/ICSharpCode.SharpZipLib.Tests/Tar/TarTests.cs
```diff
@@ -831,7 +832,38 @@ public void ExtractingCorruptTarShouldntLeakFiles()
...
+		[Category("Tar")]
+		public void StreamWithJapaneseName(int length)
+		{
+			// U+3042 is Japanese Hiragana
+			// https://unicode.org/charts/PDF/U3040.pdf
+			var entryName = new string((char)0x3042, length);
+			var data = new byte[32];
+			using(var memoryStream = new MemoryStream())
+			{
+				using(var tarOutput = new TarOutputStream(memoryStream, System.Text.Encoding.UTF8))
```
piksel, line 851:
```suggestion
				using(var tarOutput = new TarOutputStream(memoryStream, Encoding.UTF8))
```

itn3000:
Thank you for reviewing, I'd like to reflect the review results.

src/ICSharpCode.SharpZipLib/Tar/TarOutputStream.cs
itn3000: Indeed, it should be renamed to `nameEncoding`

piksel:
Regarding documentation:
Perhaps something short, like "or null for ASCII"?

Regarding Encoding.UTF8: 
That's odd? The documentation for that property explicitly warns about how it emits them:
https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.utf8?view=netframework-4.8#remarks

src/ICSharpCode.SharpZipLib/Tar/TarOutputStream.cs
piksel: Sorry, I meant the "GetAsciiBytes" method. The name is not really applicable when another encoding is used.

itn3000:
I've almost done with requested changes, please review it.

src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs
```diff
@@ -139,15 +167,15 @@ public static TarArchive CreateInputTarArchive(Stream inputStream, int blockFact
...
 
-			return new TarArchive(new TarInputStream(inputStream, blockFactor));
+			return new TarArchive(new TarInputStream(inputStream, blockFactor, nameEncoding));
 		}
-
 		/// <summary>
 		/// Create a TarArchive for writing to, using the default blocking factor
 		/// </summary>
 		/// <param name="outputStream">The <see cref="Stream"/> to write to</param>
+		/// <param name="nameEncoding"></param>
```
Numpsy, line 176: not trying to do a real review (not up to speed with the changes/requirements), but fwiw - empty doc comment for the new param?

src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs
```diff
@@ -175,6 +212,17 @@ public static TarArchive CreateOutputTarArchive(Stream outputStream)
...
 		public static TarArchive CreateOutputTarArchive(Stream outputStream, int blockFactor)
+		{
+			return CreateOutputTarArchive(outputStream, blockFactor, null);
+		}
+		/// <summary>
+		/// Create a <see cref="TarArchive">tar archive</see> for writing.
+		/// </summary>
+		/// <param name="outputStream">The stream to write to</param>
+		/// <param name="blockFactor">The blocking factor to use for buffering.</param>
+		/// <param name="nameEncoding"></param>
```
Numpsy, line 223: ditto

src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs
itn3000: There were comfirmation omission. I inserted comment.

src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs
itn3000: There were comfirmation omission. I inserted comment.

src/ICSharpCode.SharpZipLib/Tar/TarArchive.cs
piksel: Yeah, I think that the variable name here does enough to explain the purpose. I see little to no gain in writing descriptions like "The Encoding used for the name".

itn3000: Sorry for late response.
Should I leave this parameter comment empty?

piksel: It's fine

PR #372, opened 2019-08-08
Remove bad GZipException and don't loop forever on 0 reads

### LINKED ISSUE #360: Gzip decompress Unexpected EOF with small buffer + fix
### Steps to reproduce
1. Gziped 38KB file
2. Trying to ungzip it supplying into GZipInputStream.Read buffer with the size 10KB (smaller than gzipped array itself). Receive "Unexpected EOF". 
3. If to use buffer 50KB (more than gzipped array) - all works.
4. It started to work with smaller buffer when the part in GZipInputStream.Read was remarked:
image

5. Code for decompressing
```cs
public static byte[] Decompress(this byte[] data)
{
    if (data == null)
        return null;
                
    byte[] unzipped = new byte[10000];
    int readOut = 0;
    byte[] res = null;

    try
    {
        using (Stream stream = new GZipInputStream(new MemoryStream(data)))
        {
            using (MemoryStream memory = new MemoryStream())
            {
                while((readOut = stream.Read(unzipped, 0, unzipped.Length))>0)
                    memory.Write(unzipped, 0, readOut);   

                res = memory.ToArray();
                memory.Close();
            }
            stream.Close();
        }
    }
    catch
    {   
        res = null;
    }
    return res;
}
```

### Expected behavior
Unzip

### Actual behavior
Unexpected EOF

- Compiled from source, commit: 6f245b3

piksel:
PR #225 introduced a new throw that did not test the right values. The case of stream EOF is handled when attempting to fill the input buffer and `inf.RemainingInput` should not be used for this. Instead, the reason for the test getting stuck is that it attempts to read 0 bytes from the stream. Since the loop tries to read data until it can read at least one byte, it's forever stuck (since attempting to read 0 bytes never returns more than 0 bytes read).  

The use case for supplying 0 as `count` is probably slim, but at least now we're not stuck in an infinite loop.

Corr for #225
Fixes #360

### DIFF
--- a/src/ICSharpCode.SharpZipLib/GZip/GzipInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/GZip/GzipInputStream.cs
@@ -139,12 +139,10 @@ namespace ICSharpCode.SharpZipLib.GZip
 				if (inf.IsFinished)
 				{
 					ReadFooter();
-				} else if (inf.RemainingInput == 0) {
-					// If the stream is not finished but we have no more data to read, don't keep looping forever
-					throw new GZipException("Unexpected EOF");
 				}
 
-				if (bytesRead > 0)
+				// Attempting to read 0 bytes will never yield any bytesRead, so we return instead of looping forever
+				if (bytesRead > 0 || count == 0)
 				{
 					return bytesRead;
 				}


piksel:
@Numpsy Any thoughts? I've been at this for some time now, but I might have missed something crucial.

Numpsy:
I'm not sure if there are any other situations where the internal read would return 0 rather than throwing, hopefully not.

Issue #317, opened 2019-02-07
ZipFile.TestArchive seems to always return false for AES encrypted zips


Numpsy:
### Steps to reproduce

call

```
var zipFile = new ZipFile(@"AES.zip")
{
    Password = password
};

Assert.That(zipFile.TestArchive(false), Is.True, "Zip file should be valid");
```

On an AES encrypted zip file (happens with files generated with both SharpZipLib and 7-Zip).

### Expected behavior
I assume that the test should pass for a valid AES archive?

### Actual behavior
TestArchive returns false.

Looks like it fails @ https://github.com/icsharpcode/SharpZipLib/blob/master/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L1207 because the flags are different (localFlags just has the encrypted bit set, whereas entry.Flags has the StrongEncryption bit set I think).

If it gets past that, it will also fail on the compression method comparison, as entry.CompressionMethod is the compression format (Deflate), but compressionMethod is 99 (WinZipAES).

### Version of SharpZipLib
1.1.0

### Obtained from (only keep the relevant lines)
- Compiled from source, commit: 25ee644

Numpsy:
Actually, looking at the encryption bits of the spec, is it correct to set the strong encryption flag @ 

https://github.com/icsharpcode/SharpZipLib/blob/25ee6440f444aa326ee8f0b98e6223296f35fa6d/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs#L1137

?
(Seems that is supposed to be for a different encryption approach that the Winzip one)

piksel:
Yes, the only thing that the [AE-1/2](https://www.winzip.com/win/en/aes_info.html) mentions are that the encrypted flag should be set to 1 (bit zero). Strong encryption is what pkzip generally uses to refer to it's (stronger) encryption, so this seems to imply that it should not be set. The APPNOTE only specifies that 
> Strong encryption.  If this bit is set, you MUST set the version needed to extract value to at least 50 and you MUST also set bit 0.  If AES encryption is used, the version needed to extract value MUST be at least 51

Numpsy:
The logic in CreateAndInitDecryptionStream in Zipfile is currently checking for StrongEncryption being set to decide which crypto approach should be used, which seems wrong if StrongEncryption != AE-X, and only the latter is supported?

The check @ 
[...]

PR #380, opened 2019-08-25
Support AES encryption in FastZip.CreateZip

Numpsy:
Provide a means of enabling AES encryption when creating zips with FastZip (by settings the AESKeySize of the entries it creates prior to giving them to ZipOutputStream).

Notes:

- I added a new mode enum in its own file rather than in FastZip.cs (easy to merge if needed).
- I'm not sure if the encryption mode should have a 'None' option, given that the setting only applies if the password property has been set seperately?
- ~~The call to TestArchive in the unit tests is commented out because of #317~~ 

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
@@ -227,6 +227,15 @@ namespace ICSharpCode.SharpZipLib.Zip
 		}
 
 		/// <summary>
+		/// Get / set the method of encrypting entries.
+		/// </summary>
+		/// <remarks>
+		/// Only applies when <see cref="Password"/> is set.
+		/// Defaults to ZipCrypto for backwards compatibility purposes.
+		/// </remarks>
+		public ZipEncryptionMethod EntryEncryptionMethod { get; set; } = ZipEncryptionMethod.ZipCrypto;
+
+		/// <summary>
 		/// Get or set the <see cref="INameTransform"></see> active when creating Zip files.
 		/// </summary>
 		/// <seealso cref="EntryFactory"></seealso>
@@ -369,7 +378,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 			{
 				outputStream_.SetLevel((int)CompressionLevel);
 
-				if (password_ != null)
+				if (false == string.IsNullOrEmpty(password_) && EntryEncryptionMethod != ZipEncryptionMethod.None)
 				{
 					outputStream_.Password = password_;
 				}
@@ -539,6 +548,10 @@ namespace ICSharpCode.SharpZipLib.Zip
 					using (FileStream stream = File.Open(e.Name, FileMode.Open, FileAccess.Read, FileShare.Read))
 					{
 						ZipEntry entry = entryFactory_.MakeFileEntry(e.Name);
+
+						// Set up AES encryption for the entry if required.
+						ConfigureEntryEncryption(entry);
+
 						outputStream_.PutNextEntry(entry);
 						AddFileContents(e.Name, stream);
 					}
@@ -554,6 +567,26 @@ namespace ICSharpCode.SharpZipLib.Zip
 						continueRunning_ = false;
 						throw;
 					}
+				}
+			}
+		}
+
+		// Set up the encryption method to use for the specific entry.
+		private void ConfigureEntryEncryption(ZipEntry entry)
+		{
+			// Only alter the entries options if AES isn't already enabled for it
+			// (it might have been set up by the entry factory, and if so we let that take precedence)
+			if (!string.IsNullOrEmpty(Password) && entry.AESEncryptionStrength == 0)
+			{
+				switch (EntryEncryptionMethod)
+				{
+					case ZipEncryptionMethod.AES128:
+						entry.AESKeySize = 128;
+						break;
+
+					case ZipEncryptionMethod.AES256:
+						entry.AESKeySize = 256;
+						break;
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


Numpsy:
A question about naming on the encryption mode enum:

If ZipFile gets support for adding AES encrypted entries to archives then it will need some means of specifying the encryption method to use, which might mean a new property on ZipFile.
If it had such a property, the type could be the same as for FastZip, or it could be different, not sure which is best (two enums vs. the possibility that they might end up different at some point).

Basically: Is 'FastZipEncryptionMethod' a reasonable name, or would something more shareable be better?

piksel:
Hm, why would they ever need to be different? They are both working with the same spec., do you mean support-wise?

Numpsy:
Yes, support wise (though spec changes aside I don't think that would happen outside of someone implementing that PKWare strong encryption method in one and not the other, which seems unlikely in general)

src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
```diff
@@ -558,6 +571,28 @@ private void ProcessFile(object sender, ScanEventArgs e)
 			}
 		}
 
+		// Set up the encrpytion method to use for the specific entry.
```
piksel:
```suggestion
		// Set up the encryption method to use for the specific entry.
```

piksel:
I would probably go for using a common Enum.

Numpsy:
ok

Numpsy:
Thinking about my second question ("should there be 'None' encryption option") - If you were using the same enum for ZipFile then you might possibly want to be able to open a file that contains encrypted entries and add new ones which are unencrypted, which might not work if you're just basing the application of encryption on whether a password has been provided? (hence a non-option being useful).

Not sure what the best api change for ZipFile is though

src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
```diff
@@ -558,6 +571,28 @@ private void ProcessFile(object sender, ScanEventArgs e)
 			}
 		}
 
+		// Set up the encryption method to use for the specific entry.
+		private void ConfigureEntryEncryption(ZipEntry entry)
+		{
+			if (!string.IsNullOrEmpty(Password))
+			{
+				switch (EntryEncryptionMethod)
```
piksel, line 579: Codacy found an issue: [Add a 'default' clause to this 'switch' statement.](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745489)

piksel:


```diff

Issues
======
- Added 3
           

Clones removed
==============
+ test/ICSharpCode.SharpZipLib.Tests/Zip/FastZipHandling.cs  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745489&bid=18799048)

test/ICSharpCode.SharpZipLib.Tests/Zip/FastZipHandling.cs
```diff
@@ -148,8 +154,23 @@ public void Encryption()
 					ZipEntry entry = zf[0];
 					Assert.AreEqual(tempName1, entry.Name);
 					Assert.AreEqual(1, entry.Size);
-					Assert.IsTrue(zf.TestArchive(true));
+					// Assert.IsTrue(zf.TestArchive(true));
```
piksel, line 157: Codacy found an issue: [Remove this commented out code.](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745489)

test/ICSharpCode.SharpZipLib.Tests/Zip/FastZipHandling.cs
```diff
@@ -148,8 +154,23 @@ public void Encryption()
...
 					Assert.IsTrue(entry.IsCrypted);
+
+					switch (encryptionMethod)
```
piksel, line 160: Codacy found an issue: [Add a 'default' clause to this 'switch' statement.](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745489)

Numpsy:
With regards to my suggestion in #454 about creating a custom entry factory that sets the AES key size properties:

This change has a possible conflict with that, in that if you set FastZip to use ZipCrypto encryption (the default), and then set the AES key size to something non-zero in the entry factory, then FastZip would overwrite that with it's own value, Not sure how much that counts as a problem? (The way ZipOutputStream works just off the key size, and that being an int that defaults to 0, means that you can't really just put the entry factory in charge as you'd end up overriding the FastZip option with ZipCrypto in the general case.)

You could maybe make ConfigureEntryEncryption just upgrade the encryption strength or some such, but then you might just say that FastZip is for simple use cases, and this is a non-issue.

piksel:
Couldn't FastZip be modified to look at the AES key size, and not modify it if it's non-zero? Or am I missing the problem here?

Numpsy:
yes (it's just something where it needs a descision on how it should work.)

Numpsy:
updated

test/ICSharpCode.SharpZipLib.Tests/Zip/FastZipHandling.cs
```diff
@@ -187,8 +193,23 @@ public void Encryption()
 					ZipEntry entry = zf[0];
 					Assert.AreEqual(tempName1, entry.Name);
 					Assert.AreEqual(1, entry.Size);
-					Assert.IsTrue(zf.TestArchive(true));
+					// Assert.IsTrue(zf.TestArchive(true));
```
piksel, line 196:
```suggestion
					Assert.IsTrue(zf.TestArchive(true));
```
Could you leave this enabled? It would be a nice indicator for when `TestArchive` is actually fixed (although, we should probably have more tests for that).

Numpsy: sure (there might be another PR that has it disabled too, will have to remind myself)

Numpsy:
> 
> 
> I think this looks good. Does this fully enable AES encryption for FastZip, or is something else needed?

I think that might be it as far as FastZip is concerned, as all the actual work is done in ZipOutputStream, though it might need #460 being completed first to avoid it creating bad entries for folders.

Issue #379, opened 2019-08-23
Opening a tar.gz file directly from the network makes SharpZipLib to an attempt to access an already disposed Stream


koide:
I'm downloading a file from an FTP server and unzipping, untarring it on the fly, via the underlying streams. The code looks like this:
```cs
var request = WebRequest.Create("ftp://host/file.tgz");
request.Method = WebRequestMethods.Ftp.DownloadFile;
request.Credentials = new NetworkCredential("user","password");
var response = (FtpWebResponse) await request.GetResponseAsync();
using (var gzStr = new GZipInputStream(response.GetResponseStream()))
{
    gzStr.IsStreamOwner = false;
    var tar = TarArchive.CreateInputTarArchive(gzStr);
    Directory.CreateDirectory(_targetPath);
    tar.IsStreamOwner = false;
    tar.ExtractContents(_targetPath);
}
```
With the current 1.2.0 release (and 1.1.0 and 1.0.0) I get an ObjectDisposedException on the network socket stream, but it happens *after* the whole file has been extracted successfully, so it seems to be yet another exceptional case happening after finishing the file.

In my total ignorance of this code, I've tracked the issue to this line:

https://github.com/icsharpcode/SharpZipLib/blob/ab7f8c5b24c12e0656a9b8e9be55b29aeecc5593/src/ICSharpCode.SharpZipLib/GZip/GzipInputStream.cs#L122

```cs
 catch (Exception ex) when (completedLastBlock && 
    (ex is GZipException || ex is EndOfStreamException)) 
```
and, changing it to 

```cs
catch (Exception ex) when (completedLastBlock && 
    (ex is GZipException || ex is EndOfStreamException ||
     ex is ObjectDisposedException))
```

makes my specific case work successfully. I think that the server ends the connection after the file is downloaded so the stream is not there to be read from anymore.

Numpsy:
The discussion @ https://stackoverflow.com/questions/48829032/streamreader-readline-throwing-disposed-exception-rather-than-returning-null suggests that the FTP response stream closes itself when all the data is read, which could cause issues with stacked streams?

Maybe the multiple-block handling loop in GZipInputStream.Read could check stream.CanRead before trying to read a second block?

koide:
Yup, looks like it's not the server but the FtpWebRequest's FtpDataStream. 

I can confirm that doing

`while (toRead > 0 && inputStream.CanRead) `

instead of 

[...]

PR #387, opened 2019-09-15
Better handle baseStreams closing themselves unexpectedly

Numpsy:
Possible fix for #379, using the change to InflaterInputStream suggested there.

I'm not certain if its best to do the extra check in the inflator code or in gzipinputstream - putting it in the lower level code could effect the handling of Zip files as well, but i'm not certain if the issue can effect ZipInputStream in the same way as GZipInputStream.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
@@ -116,7 +116,7 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 			rawLength = 0;
 			int toRead = rawData.Length;
 
-			while (toRead > 0)
+			while (toRead > 0 && inputStream.CanRead)
 			{
 				int count = inputStream.Read(rawData, rawLength, toRead);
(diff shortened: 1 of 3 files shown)


piksel:
The question is, should we fix it? The cause of this problem is using a non-standard stream source. If it solves the problem and doesn't cause a significant performance hit, I'm okay with it. It's perhaps not an uncommon scenario.

Numpsy:
Not sure, checking if the stream can be read before reading it doesn't seem that unusual, but i'll try running the benchmarks and see if it makes a visible difference.

My first thought about this had been that you could do something inside GZipInputStream as that's the place that loops about trying to read multiple blocks and that would make it more isolated, but i haven't really tested that approach.

Numpsy:
I didn't add any GZipInputStream tests when i added the benchmark project, but fwiw the ZipInputStream benchmark doesn't show any difference with the change.

Numpsy:
@piksel I see you closed #379, does that mean that you decided that it isn't an issue that needs fixing?

piksel:
Sorry, I closed it because the problem is with `FtpDataStream` not acting like a proper `Stream`. But the change proposed is simple enough and handling it any other way is much harder. It also seems like a pretty common use case.

PR #389, opened 2019-09-23
Simplify Documentation generation

Suchiman:
It is not necessary to specify the explicit path for `<DocumentationFile>`, just setting `<GenerateDocumentationFile>` to `true` will generate the documentation file in the default locations.


Numpsy:
sounds like a good idea to me.

piksel:
From [the documentation](https://docs.microsoft.com/en-us/dotnet/csharp/codedoc):
```xml
<GenerateDocumentationFile>true</GenerateDocumentationFile>
```
This is equivalent to the following:
```xml
<DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
```
So yeah, this should indeed be exactly the same.

PR #392, opened 2019-10-31
Fix Issue #337 and Partially Fix #338

### LINKED ISSUE #337: Tar writes header names differently if the file to add is a child of the current directory.
### Steps to reproduce
1. Create an output tar archive. 
2. Set the root path to something in the same directory as the exe (so if the exe was `C:\tools\thingWithTar.exe`, set the root to `C:\tools\dirToSetAsRoot`)
3.  Attempt to add a file with the path `C:\tools\dirToSetAsRoot\fileToSend.txt`

Because the file name of the file to add has the current directory as the substring, `TarEntry.CreateEntryFromFile`, when it calls `TarEntry.GetTarFileHeader`, will set the name to

`dirToSetAsRoot\fileToSend.txt`

(this occurs on line 384 of TarEntry.cs)

Because the root path is no longer part of the name, it is not compared, and the entry added to the tar archive remains `darToSetAsRoot\fileToSend.txt`.

However, if the file was not a child of the current directory:
1. Set root to `C:\notCurDir\dirToSetAsRoot`
2. Attempt to add file `C:\notCurDir\dirToSetAsRoot\fileToSend.txt`

Then, `TarEntry.CreateEntryFromFile` names the entry: `C:\notCurDir\dirToSetAsRoot\fileToSend.txt`

Then, when written to the tar archive, the root is compared with the entry from string, which leads to the file being written as:
`fileToSend.txt`

### Expected behavior
Behavior of adding files, relative to root path, should be the same regardless if the file being added is a child of the current working directory.

### Actual behavior
The entry name differs if the file to add is a child of the current directory.

### Version of SharpZipLib
1.1.0

### LINKED ISSUE #338: Tar does not handle POSIX absolute paths correctly (trims leading '/'), breaks RootPath
Currently, TarEntry.cs, in `GetFileTarHeader`, trims all leading forward slashes.

Trying to write `/home/username/foo.txt` as an entry would to write `home/username/foo.txt`

This prevents writing absolute paths to the archive, and also breaks the RootPath string, as the entry name will no longer have the root path as a sub string (if it was set to /home/username), so it will never match when the entry is added to the archive.

If we're trying to specifically avoid UNC path names, it might be easier to just test for a leading double back slash in the original path (instead of testing for leading slashes).

### Expected behavior
RootPath should work as expected when adding absolute paths on POSIX. 

### Actual behavior
RootPath does not work as expected - file strings will never match RootPath because RootPath will still contain the leading '/', but the file entry will not. 

Additionally, absolute file paths will be broken, regardless if root path is used. 

### Version of SharpZipLib
1.1.0

iUnknwn:
Removes GetCurrentDirectory checks from TarEntry, ensuring that behavior of adding files is consistent, regardless of the current assembly location (resolves #337).

Also removes the check for leading slashes, which prevents absolute paths on POSIX systems from being properly stored. This partially fixes #338, but not completely, as it also means UNC paths are unblocked (an unintended side effect of the fix). 


piksel:
It should be possible to fix this without allowing absolute paths. If we really want to add support for them, they should have to be explicitly be turned on since they are usually not intended or wanted (GNU tar strips them if not `-P` is supplied). The fact that this also allows creating files UNC paths makes it even less of a good idea. Whether intentional or not, the consequences of such entries could cause all kinds of issues.

The name "cleaning" should of course be done in the same name on both sides of the equality check.

src/ICSharpCode.SharpZipLib/Tar/TarEntry.cs
```diff
@@ -404,14 +408,6 @@ public void GetFileTarHeader(TarHeader header, string file)
 
 			name = name.Replace(Path.DirectorySeparatorChar, '/');
 
-			// No absolute pathnames
```
piksel: Not a viable solution.

src/ICSharpCode.SharpZipLib/Tar/TarEntry.cs
```diff
@@ -379,10 +379,14 @@ public void GetFileTarHeader(TarHeader header, string file)
 			string name = file;
 
 			// 23-Jan-2004 GnuTar allows device names in path where the name is not local to the current directory
+			// 04-Apr-2019 Removed this block, as it creates different behavior based the location of the executing
+			//             assembly.
+			/*
```
piksel: The purpose of this block is to do exactly that. While I do agree that this may not belong here, just removing it will just break code relying on this behaviour.

piksel:


```diff

Complexity decreasing per file
==============================
+ src/ICSharpCode.SharpZipLib/Tar/TarEntry.cs  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745487&bid=18799046)

PR #402, opened 2019-12-14
Don't convert entry.Name twice when accessing updateIndex

### LINKED ISSUE #82: Can not delete directory from zip archive bug
ICSharpCode.SharpZipLib version - 0.86.0.518.

Code to demonstrate the bug:

``` csharp
namespace SharpZipLibDeleteFolderBug
{
    using ICSharpCode.SharpZipLib.Zip;

    using NUnit.Framework;

    [TestFixture]
    public class SharpZipLibDeleteFolderBugDemo
    {
        [Test]
        public void Bug()
        {
            var zipFile = ZipFile.Create("SomeFile.zip");
            CreateDirectory(zipFile, "SomeDirectory");

            DeleteDirectory(zipFile, "SomeDirectory");
        }

        private void DeleteDirectory(ZipFile zipFile, string directoryName)
        {
            zipFile.BeginUpdate();

            var zipEntry = zipFile.GetEntry(directoryName + "/");
            Assert.That(zipEntry, Is.Not.Null); // passed - directory was successfully create

            zipFile.Delete(zipEntry); // !!! throws ICSharpCode.SharpZipLib.Zip.ZipException : Cannot find entry to delete

            zipFile.CommitUpdate();            
        }

        private static void CreateDirectory(ZipFile zipFile, string directoryName)
        {
            zipFile.BeginUpdate();
            zipFile.AddDirectory(directoryName);
            zipFile.CommitUpdate();
        }
    }
}
```

The buggy method is ZipFile.FindExistingUpdate:

``` csharp
        int FindExistingUpdate(ZipEntry entry)
        {
            int result = -1;
            string convertedName = GetTransformedFileName(entry.Name); // !!! BUG is here

            if (updateIndex_.ContainsKey(convertedName)) {
                result = (int)updateIndex_[convertedName];
            }

                        ...
            return result;
        }
```

Method FindExistingUpdate ignores the fact that entry can be not only file entry but also a directory entry. Right implementation should be:

``` csharp
        int FindExistingUpdate(ZipEntry entry)
        {
            int result = -1;
            string convertedName;
            if (entry.IsDirectory)
                convertedName = GetTransformedDirectoryName(entry.Name);
            else 
                convertedName = GetTransformedFileName(entry.Name);

            if (updateIndex_.ContainsKey(convertedName)) {
                result = (int)updateIndex_[convertedName];
            }

                        ...
            return result;
        }
```

Method ZipFile.Delete(string fileName) also has the same bug.

AqlaSolutions:
I found out that deleting empty directory entry from ZipFile and then adding it back causes an exception while trying to add to `updateIndex` dictionary - because checking for existing element uses transformed `entity.Name` while adding uses its original value. Also directory name gets transformed as file name and ending slash gets removed therefore those too values are not same especially in directory case.
`Entry.Name` is already transformed file name so it shouldn't be transformed twice. So I replaced all such accesses to use entry.Name or transformed string-filename.

Fixes #82

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -1620,7 +1620,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		{
 			contentsEdited_ = true;
 
-			int index = FindExistingUpdate(update.Entry.Name);
+			int index = FindExistingUpdate(update.Entry.Name, true);
 
 			if (index >= 0)
 			{
@@ -2529,13 +2529,9 @@ namespace ICSharpCode.SharpZipLib.Zip
 		private int FindExistingUpdate(ZipEntry entry)
 		{
 			int result = -1;
-			string convertedName = entry.IsDirectory
-				? GetTransformedDirectoryName(entry.Name)
-				: GetTransformedFileName(entry.Name);
-
-			if (updateIndex_.ContainsKey(convertedName))
+			if (updateIndex_.ContainsKey(entry.Name))
 			{
-				result = (int)updateIndex_[convertedName];
+				result = (int)updateIndex_[entry.Name];
 			}
 			/*
 						// This is slow like the coming of the next ice age but takes less storage and may be useful
@@ -2553,11 +2549,11 @@ namespace ICSharpCode.SharpZipLib.Zip
 			return result;
 		}
 
-		private int FindExistingUpdate(string fileName)
+		private int FindExistingUpdate(string fileName, bool isEntityName = false)
 		{
 			int result = -1;
 
-			string convertedName = GetTransformedFileName(fileName);
+			string convertedName = !isEntityName ? GetTransformedFileName(fileName) : fileName;
 
 			if (updateIndex_.ContainsKey(convertedName))
 			{
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -1620,7 +1620,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		{
 			contentsEdited_ = true;
 
-			int index = FindExistingUpdate(update.Entry.Name, true);
+			int index = FindExistingUpdate(update.Entry.Name, isEntityName: true);
 
 			if (index >= 0)
 			{
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -1620,7 +1620,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		{
 			contentsEdited_ = true;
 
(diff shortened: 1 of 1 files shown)


Numpsy:
is this covering similar ground to #309? (haven't looked at the issue, just a query as they look to be changing similar things)

piksel:
Yes, that PR was supposed to fix that issue, but it's been stale for almost a year.

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -1620,7 +1620,7 @@ private void AddUpdate(ZipUpdate update)
 		{
 			contentsEdited_ = true;
 
-			int index = FindExistingUpdate(update.Entry.Name);
+			int index = FindExistingUpdate(update.Entry.Name, true);
```
piksel: I try to add labels for all booleans to make the code more descriptive (without relying on an IDE).
```suggestion
			int index = FindExistingUpdate(update.Entry.Name, isEntityName: true);
```

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -2553,11 +2549,11 @@ private int FindExistingUpdate(ZipEntry entry)
 			return result;
 		}
 
-		private int FindExistingUpdate(string fileName)
+		private int FindExistingUpdate(string fileName, bool isEntityName = false)
```
piksel: Is this supposed to be `isEntryName`? Not sure why the word "entity" is used here?

AqlaSolutions: You are right, it should be `isEntryName`

piksel:
```suggestion
		private int FindExistingUpdate(string fileName, bool isEntryName = false)
		{
			int result = -1;

			string convertedName = GetTransformedFileName(fileName);
			string convertedName = !isEntryName ? GetTransformedFileName(fileName) : fileName;
```

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -2553,11 +2549,11 @@ private int FindExistingUpdate(ZipEntry entry)
...
 		{
 			int result = -1;
 
-			string convertedName = GetTransformedFileName(fileName);
+			string convertedName = !isEntityName ? GetTransformedFileName(fileName) : fileName;
```
piksel:
```suggestion
			string convertedName = !isEntryName ? GetTransformedFileName(fileName) : fileName;
```

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -1620,7 +1620,7 @@ private void AddUpdate(ZipUpdate update)
 		{
 			contentsEdited_ = true;
 
-			int index = FindExistingUpdate(update.Entry.Name);
+			int index = FindExistingUpdate(update.Entry.Name, isEntityName: true);
```
piksel:
```suggestion
			int index = FindExistingUpdate(update.Entry.Name, isEntryName: true);
```

Issue #403, opened 2019-12-20
Zip64 signature found in "standard" file. As a result of "357. LocateBlockWithSignature returns false position.


JanneSaare:
Got a file, ziped by SharpZipLib, that obviously has the signature of ZIP64 some where inside. When unzipping the LocateBlockWithSignature returns a position. This leads to End of Stream error in ZIP64 processing.
I can not send an example file (GDPR)
### Steps to reproduce
1. ZIP archive that in its bitsream has by random the ZIP64 signature.
2. Try open archive new ZipFile(SelectedCaseFilePath)
3. LocateBlockWithSignature returns a position resulting in wrong execution path
4. To debug:
 If, in debug, force "standard" decompression the file opens OK.
// #357 - always check for the existance of the Zip64 central directory.
long locatedZip64EndOfCentralDir = LocateBlockWithSignature(ZipConstants.Zip64CentralDirLocatorSignature, locatedEndOfCentralDir, 0, 0x1000); //Returns a position
			if (locatedZip64EndOfCentralDir < 0)
>Force execution>>			{
				if (requireZip64)

### Expected behavior
A ZipFile object

### Actual behavior
End of Stream error.

### Version of SharpZipLib
1.2.0

### Obtained from (only keep the relevant lines)
- Compiled from source, commit: 56cbe99 on 9 Aug
- Downloaded from GitHub
- Package installed using NuGet

Numpsy:
hmm, I hadn't noticed when I made that previous change for zip64 handling, but looking now I wonder if the 3rd and 4th parameters to LocateBlockWithSignature are correct?

If it's looking for the zip 64 end of central directory locator block and the definition of that is

```
zip64 end of central dir locator 
      signature                       4 bytes  (0x07064b50)
      number of the disk with the
      start of the zip64 end of 
      central directory               4 bytes
      relative offset of the zip64
      end of central directory record 8 bytes
      total number of disks           4 bytes
```

The setting minimumBlockSize to 0 and maximumVariableData to 0x1000 seems wrong? (i.e. should it be 20/0?)

piksel:
@Numpsy It should, but that would only speed up the search.

[...]

Issue #375, opened 2019-08-12
FastZip.CreateZip() , filter "*.o" will show all files ,filter ".so" will show "*.*so*",it's bad


hygyh:
### Expected behavior
1.the directory has files,ex:xxx.o,xxx.so,xx.soc,xx.oso
2.aFileFiter = ".so" 
3.new FastZip().CreateZip(aZipedFileName, aFilePath, true, aFileFiter, aDirectoryFiter);
4.result will show xx.so ,xx.soc ,xx.oso

Numpsy:
I think those filters are supposed to be regular expressions, so those cases might be  "anything that includes 'o'" and "anything that includes 'so'"?

(/regex knowledge is extremely basic)

piksel:
Indeed!

https://regex101.com/r/0dG0vG/1

piksel:
Closing this as it seems to have been answered.

PR #408, opened 2020-01-06
When searching for the Zip64 end of central directory locator, pay attention to its fixed size.

Numpsy:
#403 / #375 - the Zip64 end of central directory locator has a fixed 20 byte length, but the calls to LocateBlockWithSignature were specifying a fixed size of 0 and a dynamic size of 0x1000, which meant that the signature bytes could incorrectly be matched in a location other than where they should be (20 bytes prior to the end of central directory signature)

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipConstants.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipConstants.cs
@@ -345,6 +345,11 @@ namespace ICSharpCode.SharpZipLib.Zip
 		[Obsolete("Use CryptoHeaderSize instead")]
 		public const int CRYPTO_HEADER_SIZE = 12;
 
+		/// <summary>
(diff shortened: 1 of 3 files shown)


Numpsy:
(Might also be worth adding an extra unit test with a file crafted to fall over 403)

Numpsy:
Added a simple unit test that fails with the existing code and passes with the change

Numpsy:
I hadn't considered a different structure in self extractors, but I don't know enough about the differences to say either way.

piksel:
Well, one of the interesting parts about the zip format is that, compared to most other file formats, it is basically read from the end. The "true" entry point into the data structure is the "end of central directory record". This means that you can prepend other data before the file entries, like for example a windows PE binary or an image. This means that a file can be both a valid .exe and a valid .zip at the same time.

PR #411, opened 2020-01-10
fix ZipStrings.UseUnicode

theKBro:
use CodePage property instead of backing field so UseUnicode is set correct even if codePage is auto

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs
@@ -80,17 +80,17 @@ namespace ICSharpCode.SharpZipLib.Zip
 		{
 			get
 			{
-				return codePage == Encoding.UTF8.CodePage;
+				return CodePage == Encoding.UTF8.CodePage;
 			}
 			set
 			{
 				if (value)
 				{
-					codePage = Encoding.UTF8.CodePage;
+					CodePage = Encoding.UTF8.CodePage;
 				}
 				else
 				{
-					codePage = SystemDefaultCodePage;
+					CodePage = SystemDefaultCodePage;
 				}
 			}
 		}


piksel:
I don't think this does what you think it does. Could you explain the issue you are trying to solve?

theKBro:
ZipStrings.codePage is initialized with AutomaticCodePage(-1) and as the comment says it is intended to be good citizens and default to UTF-8.
So ZipStrings.CodePage.get returns Encoding.UTF8.CodePage.
But ZipStrings.UseUnicode.get returns false in this case which is not correct from my perspective.
Now ZipStrings.UseUnicode.get returns true in this case.

JCH2k:
The issue was: We create simple ZIP files from a directory with all Zip entries using Deflate and AES. The name is just taken from the original (unzipped) files (passed as a .NET string).
Now unicode chars like ä, ö, ü, ß were not encoded correctly in all entry names (also shown by 7-zip). While debugging, we discovered that ZipStrings.UseUnicode was always returning false, which means unicode is not used at all.

As @theKBro states 2 things make no sense to us in the current implementaion:
- `ZipStrings.CodePage` returns `UTF8` but `ZipStrings.UseUnicode` returns false
- The comment on the `codePage` field clearly says that the default should be UTF8, but eventually all our file names were encoded in IBM437.

The change in this PR makes encoding use UTF8 while decoding still uses the system's code page unless code page is set by flags, which makes sense to me.

Are we overlooking something?

piksel:
The settings in ZipStrings relates to how the bytes are written/interpreted in zip archives. When creating a new archive, you should set the `ZipEntry.IsUnicodeText` property on the items. This way, the entry will have the Unicode bit set and the file names and comments will be written using UTF-8 bytes.

The only reason to override the ZipStrings is when you want to read/write files that do not comply with the ZIP format.

The reason for "UseUnicode" returning false is because it has not been overriden. You CAN do that, by changing the UseUnicode property, but it will just create a ZIP file that will have corrupt file names in most zip programs.

The comment you mention should be removed as it refers to the change that caused SharpZipLib to create non-conforming zip files and what later reverted. Wanting UTF-8 everywhere is nice, but you cannot just stop following the file format. Instead, what we should do, is to set the `ZipEntry.IsUnicodeText` to `true` as the default for new archives.

JCH2k:
Yes, setting the default value of `ZipEntry.IsUnicodeText` to `true` would also remove the confusion and make the library easier to use.
Should I create an Issue for that?

piksel:
@JCH2k sure! The whole ZipStrings class is utterly confusing and should be rewritten as a instanced object on the Input/Output streams instead, with better descriptions and naming.
But I think that would be a good start, and also removing (or maybe moving to the IsUnicodeText property?) the aforementioned comment that isn't actually true anymore.

piksel:
Btw. could either you, or @MortenJoergensen supply a short snippet of how you used the `ZipStrings.UseUnicode` property? Since you both had the same idea of how it ought to be used, there's clearly a risk of it happening to others. It will give me a better idea of how to update the documentation as a short term solution.

MortenJoergensen:
I think the problem comes from not doing anything with ZipStrings.UseUnicode at all. I just did a simple new FastZip() (with some events as well), and then expected calling CreateZip to work out of the box (which it did, more or less). Debugging a little revealed that ZipStrings.UseUnicode returned false, and comparing the UseUnicode getter with the CodePage getter above it, made me conclude that it was supposed to default to using unicode, if not explicitly set. Doing a ZipString.UseUnicode = true works just fine. However, I agree with your comment about rewriting ZipStrings as an instanced object. If creating a FastZip object or calling CreateZip somehow forces the caller to consider which flags and properties to set, running with simple defaults becomes less likely - I don't know if going that far was part of your idea for rewriting.

Issue #162, opened 2017-01-14
Support for multi-stream BZip2


FabriceBoyer:
When trying to decompress BZ2 wikipedia dumps (https://dumps.wikimedia.org/enwiki/20170101), I reach an unexpected EndOfFile exception

This is apparently linked to the usage of parallel compression when creating the bzip2 file. More informations about this feature and a possible workaround are present at the following link : 
http://chaosinmotion.com/blog/?p=723

Would it be possible to add this feature to SharpZipLib so we can uncompress these very large files only in memory ?

Regards,
Fabrice

PR #417, opened 2020-02-01
Add multistream support to BZip2

### LINKED ISSUE #413: Getting error "Index was outside the bounds of the array" when writing uncompressed data
### Steps to reproduce
1. Have a .txt.bz2 file that has more header columns than actual values
Example:
| first_name | last_name | middle_name |
| John | Jones |  |

### Expected behavior
When running `BZip2.Decompress(fileToDecompressAsStream, decompressedStream, true);`, it should obviously be writing the data from the `.bz2` file to the `.txt` file.

### Actual behavior
I'm getting an error stating `Index was outside the bounds of the array`

I know that this issue is more of a "bad data" problem, rather than a problem with SharpZipLib. However, I was hoping you could help with finding a solution. Ideally, I would like to decompress the .bz2 file, and either remove the extra header column or give all the rows a `NULL` value. But I can't find a way to do this. Any help you could provide would be very much appreciated!

### Version of SharpZipLib
1.2.0

piksel:
Fixes #413 and #162

### DIFF
--- a/src/ICSharpCode.SharpZipLib/BZip2/BZip2.cs
+++ b/src/ICSharpCode.SharpZipLib/BZip2/BZip2.cs
@@ -15,7 +15,8 @@ namespace ICSharpCode.SharpZipLib.BZip2
 		/// <param name="inStream">The readable stream containing data to decompress.</param>
 		/// <param name="outStream">The output stream to receive the decompressed data.</param>
 		/// <param name="isStreamOwner">Both streams are closed on completion if true.</param>
-		public static void Decompress(Stream inStream, Stream outStream, bool isStreamOwner)
+		/// <param name="multiStream">Whether to attempt to continue reading after stream end.</param>
+		public static void Decompress(Stream inStream, Stream outStream, bool isStreamOwner, bool multiStream)
 		{
 			if (inStream == null || outStream == null)
 			{
@@ -24,7 +25,7 @@ namespace ICSharpCode.SharpZipLib.BZip2
 
 			try
 			{
-				using (BZip2InputStream bzipInput = new BZip2InputStream(inStream))
+				using (BZip2InputStream bzipInput = new BZip2InputStream(inStream, multiStream))
 				{
 					bzipInput.IsStreamOwner = isStreamOwner;
 					Core.StreamUtils.Copy(bzipInput, outStream, new byte[4096]);
@@ -39,6 +40,16 @@ namespace ICSharpCode.SharpZipLib.BZip2
 				}
 			}
 		}
+
+		/// <summary>
+		/// Decompress the <paramref name="inStream">input</paramref> writing
+		/// uncompressed data to the <paramref name="outStream">output stream</paramref>
+		/// </summary>
+		/// <param name="inStream">The readable stream containing data to decompress.</param>
+		/// <param name="outStream">The output stream to receive the decompressed data.</param>
+		/// <param name="isStreamOwner">Both streams are closed on completion if true.</param>
+		public static void Decompress(Stream inStream, Stream outStream, bool isStreamOwner)
+			=> Decompress(inStream, outStream, isStreamOwner, true);
 
 		/// <summary>
 		/// Compress the <paramref name="inStream">input stream</paramref> sending
--- a/src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
@@ -87,16 +87,22 @@ namespace ICSharpCode.SharpZipLib.BZip2
 		private int i2, j2;
 		private byte z;
 
+		private bool multiStream;
+		private long processedLastStream;
+
 		#endregion Instance Fields
 
 		/// <summary>
 		/// Construct instance for reading from stream
 		/// </summary>
 		/// <param name="stream">Data source</param>
-		public BZip2InputStream(Stream stream)
+		/// <param name="multiStream">Whether to continue reading streams after the first one</param>
+		public BZip2InputStream(Stream stream, bool multiStream)
 		{
-			if (stream == null)
-				throw new ArgumentNullException(nameof(stream));
+			baseStream = stream ?? throw new ArgumentNullException(nameof(stream));
+			this.multiStream = multiStream;
+
+
 			// init arrays
 			for (int i = 0; i < BZip2Constants.GroupCount; ++i)
 			{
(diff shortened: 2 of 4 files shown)


Numpsy:
(fixes #413 rather than 276?)

piksel:
Gah. Github! Instead of using the number I actually typed, it auto-completed to the first search result that started with the number (413).

Numpsy:
Not sure if this would be considered a real issue or not, but in case:

Does the usage of the Length property on the baseStream have any ramifications on the supported types of input stream?

Example: Say that I have a Zip file that contains a BZip2 file. With the current code I appear to be able to open a BZip2InputStream on the stream that is returned from ZipFile.GetInputStream and read the data from it, but with this change I get

```
System.NotSupportedException : InflaterInputStream Length is not supported
```

I'm not sure off hand is BZip2InputStream is inteded to support such streams though? (might possibly also happen if you tried it with a network stream or some such as well)

philippk80:
In which release this will be available?

piksel:
This will probably be in the next release. I have been super busy and neglected this unfortunately. Hopefully I will get time this weekend to fix current issues.

src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
```diff
@@ -286,7 +294,17 @@ public override int ReadByte()
 		{
 			if (streamEnd)
 			{
-				return -1; // ok
+				long bytesProcessed = baseStream.Position - processedLastStream;
```
piksel:
```suggestion
                if(!multiStream) {
                    return -1; // no multistream support, so just end it here
                }
                
				long bytesProcessed = baseStream.Position - processedLastStream;
				long bytesLeft = baseStream.Length - baseStream.Position;
				if (bytesProcessed > 0 && bytesLeft > 0)
				{
					processedLastStream = bytesProcessed;
					Initialize();
				}
				else
				{
					return -1; // ok
				}
```

src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
```diff
@@ -105,14 +111,16 @@ public BZip2InputStream(Stream stream)
...
 			Initialize();
-			InitBlock();
-			SetupBlock();
 		}
 
+		/// <summary>
+		/// Construct instance for reading from stream
+		/// </summary>
+		/// <param name="stream">Data source</param>
+		public BZip2InputStream(Stream stream) : this(stream, true) { }
```
piksel:
```suggestion
		public BZip2InputStream(Stream stream) : this(stream, stream.CanSeek) { }
```

src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
```diff
@@ -286,7 +294,17 @@ public override int ReadByte()
...
+				long bytesLeft = baseStream.Length - baseStream.Position;
+				if (multiStream && bytesProcessed > 0 && bytesLeft > 0)
+				{
+					processedLastStream = bytesProcessed;
+					Initialize();
+				}
+				else
+				{
+					return -1; // ok
+				}
```
piksel:
```suggestion
                if(!multiStream) {
                    return -1; // no multistream support, so just end it here
                }
                
				long bytesProcessed = baseStream.Position - processedLastStream;
				long bytesLeft = baseStream.Length - baseStream.Position;
				if (bytesProcessed > 0 && bytesLeft > 0)
				{
					processedLastStream = bytesProcessed;
					Initialize();
				}
				else
				{
					return -1; // ok
				}
```

src/ICSharpCode.SharpZipLib/BZip2/BZip2.cs
```diff
@@ -40,6 +41,16 @@ public static void Decompress(Stream inStream, Stream outStream, bool isStreamOw
...
 
+		/// <summary>
+		/// Decompress the <paramref name="inStream">input</paramref> writing
+		/// uncompressed data to the <paramref name="outStream">output stream</paramref>
+		/// </summary>
+		/// <param name="inStream">The readable stream containing data to decompress.</param>
+		/// <param name="outStream">The output stream to receive the decompressed data.</param>
+		/// <param name="isStreamOwner">Both streams are closed on completion if true.</param>
+		public static void Decompress(Stream inStream, Stream outStream, bool isStreamOwner)
+			=> Decompress(inStream, outStream, isStreamOwner, true);
```
piksel:
```suggestion
			=> Decompress(inStream, outStream, isStreamOwner, inStream.CanSeek);
```

piksel:
@Numpsy I made suggestions that should make it compatible with non-seekable streams. It will not access Stream.Length if `multistream` is not enabled, and it defaults to enable it only if the input stream supports seeking.

src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
Numpsy: seems reasonable

src/ICSharpCode.SharpZipLib/BZip2/BZip2.cs
Numpsy: I'm not certain if there aren't any streams that support seeking but don't allow access to their length, but maybe that's too much of an unusual situation to worry about, and as long as the functionality can be disabled that should cover the situation if needed.

piksel: I don't know if seeking can even be achieved if you don't allow Length (since the SeekOrigin parameter of Seek call can be set to "End".
The Stream documentation basically thinks of them as mutually inclusive:
https://docs.microsoft.com/en-us/dotnet/standard/io/?view=netframework-4.8#streams

Numpsy: That's why I said too unusual to worry about (it might not make sense, but that doesn't guarantee it won't happen!).
Not a reason to avoid making the change though

piksel:


```diff

Issues
======
+ Solved 1
- Added 1
           

Complexity decreasing per file
==============================
+ src/ICSharpCode.SharpZipLib/BZip2/BZip2Constants.cs  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745484&bid=18799043)

src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
```diff
@@ -87,16 +87,22 @@ during decompression.
 		private int i2, j2;
 		private byte z;
 
+		private bool multiStream;
```
piksel, line 90: Codacy found an issue: [Make 'multiStream' 'readonly'.](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745484)

PR #419, opened 2020-02-09
Use github workflows for CI

### LINKED ISSUE #434: Q. Multi-targeted unit tests?
Think i've wondered this before after seeing differences between the way that .NET and .NET Core handle validation of characters in paths, but thinking about it again in regards to code page handling with #427:

The unit tests are currently run as .NET Core 2.0 only, but they could be made to multi-target both some version of Core, and .NET 4.5 as well (to match the targets on the lib itself) ?

VS 2019 and the dotnet command line tools will handle that ok (will run under multiple frameworks and display multiple results sets), though the test runner in VS 2017 will only run one of the targets.

piksel:
To enable better feedback and integration, this PR adds Github Actions to build and test merges into master and pull requests.
This includes running on `windows`, `macOS` and `linux` platforms and both on .NET Framework (on windows) and .NET Core.

Fixes #434

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -1,5 +1,8 @@
 name: Build master on push
 
+env:
+  PUBLISH_DEV_PACKS: disabled
+
 on:
   push:
     branches: [ master ]
@@ -25,7 +28,7 @@ jobs:
       if: matrix.libframework == 'netstandard2'
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 2.2.108
+        dotnet-version: '3.1.x'
 
     - name: Build library
       run: dotnet build -c ${{ matrix.configuration }} -f ${{ matrix.libframework }} src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -42,7 +45,7 @@ jobs:
         
 
   Package:
-    if: 1 == 2
+	if: env.PUBLISH_DEV_PACKS != 'disabled'
     needs: [BuildAndTest]
     runs-on: windows-latest
     env:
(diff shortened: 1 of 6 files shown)


Numpsy:
Not directly a question about the CI system, but related and something I've wondered before:

What, at this point, does the manual TestBootstrapper app give you as compared to just using the 

```
dotnet test
```

command on the test project?
The test command should be operable cross platform, and handle multi-targeted tests and such, which seems useful for the test builds?

piksel:
If I recall correctly, it's the report. But whatever reason there used to be, it might not be valid anymore.

Numpsy:
I believe there is an appveyor logger for dotnet test, and looking around I see https://github.com/Tyrrrz/GitHubActionsTestLogger, which sounds like it might be useful for displaying test failures in the Actions output?

Numpsy:
The [actions documentation](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners) suggests that there is a macOS runner available for builds - might be use to test with that on top on Windows and Linux?

Numpsy:
is the 'Build and Test PR' script running the tests as part of the 'build' task, and then again as part of 'test' task?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,79 @@
...
+          - os: macos-latest
+            libframework: netstandard2
+    steps:
+    - uses: actions/checkout@v2
+      
+    - name: Setup .NET Core
+      if: matrix.libframework == 'netstandard2'
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 2.2.108
```
Numpsy, line 28: Is 2.2.108 the appropriate version to use at this point? (I mean, I think it's one of the versions that's out of support, and I think the other PR for the nullability changes would require a v3 SDK for C#8 support)

piksel: Yeah, I think using the latest stable would be more appropriate

Issue #423, opened 2020-02-12
Test WriteZipStreamWithNoCompression Failed


workgroupengineering:
### Steps to reproduce
1. Run test

### Expected behavior
testing must not fail

### Actual behavior
Test throw "No entry matching source file \"{0}\" found in archive, found \"{1}\""

### Version of SharpZipLib
cuurent SharpZipLib/master
### Obtained from (only keep the relevant lines)
- Compiled from source, commit: 803b4a2
- Downloaded from GitHub

Numpsy:
Related to the discussion in #341 - I'm not sure whether the name transforming/cleaning should be done in ZipOutputStream, rather than always being left up to the caller.

Numpsy:
should be fixed now?

piksel:
Yeah, I moved this to it's own test in #502 and forgot to link this issue!

Since #497 the underlying cause should be fixed as well.

PR #424, opened 2020-02-12
Fixed Issue #423

workgroupengineering:


piksel:
As mentioned by @Numpsy, we have to make a decision about how to correctly handle this. It might very well be that this is the end result, but if so, other places need to be update as well.

PR #381, opened 2019-08-29
[WIP] AES decryption support in ZipInputStream

Numpsy:
I started looking at this and then haven't had time to try to finish it yet, so here it is in case anyone has any ideas.

Tne basic decryption seems to be ok, but the auth code checking is disabled as it doesn't work. I think the issue is the way that inflater input buffer runs all its data through the crypto transform rather than just the part which is actually encrypted, so the auth code calculation in the AES transform gets the wrong result.
Not sure of the best way to handle it - might need changes in the input buffer, or a different approach to calculating the value?

The changes also have a couple of more minor changesin seperate commits, in case they're useful on their own.

_I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the SharpZipLib open source product (the "Contribution"). My Contribution is licensed under the MIT License._

PR #407, opened 2019-12-31
Allow Descriptor flag for Stored method

piksel:
Fixes #384 

Tests needs fixing, and some more tests needs to be added. There might also be some Descriptor-related code that is disabled due to a Deflate guard.

_I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the SharpZipLib open source product (the "Contribution"). My Contribution is licensed under the MIT License._

PR #436, opened 2020-04-05
In ZipInputStream, store the entry compression method as a CompressionMethod, instead of storing it as an int and casting it in multiple places

Numpsy:
A piece of #381 broken out into its own PR - rather than storing the compression method of the current entry as an int and casting in multiple places, store it as a CompressionMethod with one cast and then use it directly elsewhere.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
@@ -73,7 +73,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		private ZipEntry entry;
 
 		private long size;
-		private int method;
+		private CompressionMethod method;
 		private int flags;
 		private string password;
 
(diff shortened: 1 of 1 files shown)


piksel:
Yeah, I did this for the bitflags in #407. There might be a performance cost to this, but for most cases I think it's compile-time only.

piksel:
That title tho. :)

Numpsy:
I'd hope there wasn't a runtime perf cost, but i'm not certain.

The bigger change I made in #381 actually removed the local method member and used the property on the stored ZipEntry instead, which would be more likely to have an impact, but which I think I did to handle the WinZipAes case as well (where the ZipEntry property replaces the '99' AES method with the real compression method)

piksel:
Yeah, normally I would say that worrying about compiler/runtime internals is a waste of time, but this project is rather special in that seemingly irrelevant changes might induce huge a performance impact when the code is ran as many times a second as the resources can handle.

Issue #439, opened 2020-04-08
Adding file to archive when existing entry has AES encryption removes encryption?


jlficken:
### Steps to reproduce
1. Create zip file with password protected zip entry that has AES encryption
2. Follow this tutorial - https://github.com/icsharpcode/SharpZipLib/wiki/Updating
3.

### Expected behavior
Existing entry should still have AES encryption

### Actual behavior
AES encryption has been removed from the existing entry

### Version of SharpZipLib
1.2.0.246

### Obtained from (only keep the relevant lines)
- Package installed using NuGet

Numpsy:
I think that what happens in the simple add-new-file case might be that the encrypted entry data is copied across into the updated file, but when the central directory records are updated, it simply writes the wrong value for the compression method (so the entry data does remain encrypted, but the compression method is set to the actual compression method instead of 99)

The line @

https://github.com/icsharpcode/SharpZipLib/blob/fed3bd219f8bd2bac5287b6217b9c4c384bed35f/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L2217

should perhaps be

```
WriteLEShort((byte)entry.CompressionMethodForHeader);
```

ditto for

https://github.com/icsharpcode/SharpZipLib/blob/fed3bd219f8bd2bac5287b6217b9c4c384bed35f/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L2107

for a start?

Numpsy:
@jlficken would you be able to see if the issue you're seeing is resolved in the latest code? (the latest CI builds for the master branch should include the fix)

jlficken:
I'll check it out in the morning and report back.

Thanks!

jlficken:
Dumb question but if I'm using NuGet in VS2015 what's the easiest way to test this or would I have to wait until it's released.  I haven't started using GitHub or NuGet until recently so I'm still trying to figure  things out.

Numpsy:
You could try adding the appveyor CI package feed to your solution.
If you create a file called 'NuGet.config' in your solution folder, and add

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
	<packageSources>
		<add key="szl" value="https://ci.appveyor.com/nuget/sharpziplib-cinugets" />
	</packageSources>
</configuration>
```

[...]

PR #440, opened 2020-04-10
Use ZipEntry.CompressionMethodForHeader when writing the header entries

Numpsy:
When writing the local/central directory entries in ZipFile, use the CompressionMethodForHeader property of ZipEntry instead of the CompressionMethod property, in order to handle the case where the entry method is WinZipAes.

Add a couple of extra unit tests, which both try to verify using SharpZipLib locally (#317 aside) and with 7-Zip.

refs #439 


Numpsy:
The CompressionMethodForHeader property shouid only differ from the CompressionMethod property for AES encrypted entries, and as you can't presently add those via ZipFile the change should only effect updates to existing files that contain AES encrypted entries.

I don't think there are any situations where this would break anything, given that the current code already breaks such entries when the file is updated?

piksel:
No, it shouldn't afaik.

PR #442, opened 2020-04-11
Add self-extracting example project

piksel:


piksel:
This should be simplified as a "sdk" project before merging

piksel:
I am getting quite skeptical about this... The concept isn't that complicated, you can just append a .ZIP to the end of an executable, which can then treat itself as a zip file and extract the contents. This works because zipfiles **end** with the "header", and .EXE files **start** with their header.

The problem with using SharpZipLib for this is that if you reference either the .csproj (as this sample currently does), or the nuget package, you get an additional .dll in the output folder. What is the point of a self-extrating *pair* of files? But OK, you can use ilmerge to combine them into a single assembly, but then there's the next problem... The overhead is ~320kb. Compare that to using something like 7z SFX, which is ~200kb with much greater performance and compression ratio...

piksel:
Closing this due to the above stated reasons.

PR #443, opened 2020-04-11
[WIP] Add AES encryption support to ZipFile

Numpsy:
A work-in-progress attempt to add AES encryption support to ZipFile.

Notes:

There is currently no way to tell ZipFile what encryption algorithm to use when adding entries, so the only way to add an AES entry is with the variant of Add() that takes a ZipEntry (and set the AESKeySize property on it). Perhaps needs to be considered together with #380.

I copied the AddExtraDataAES function from the zip stream code, it would be better being shared somewhere. (an extension method on ZipExtraData maybe?)

It tweaks the logic in GetOutputStream() to try to ensure that the crypto streams returned from it are disposed - perhaps could be a separate PR as it might be a sort of existing issue with it not disposing streams when it should?

The Write() overload in ZipAESEncryptionStream takes the same approach as ZipOutputStream and calls the crypto transform TransformBlock() on each write. I'm not sure if that is technically correct, or if it should store the data and transform it in complete blocks? (or perhaps it doesn't matter with the way the transform is implemented)

Doesn't yet add any new unit tests (would likely end up conflicting with other PRs in the same area, could maybe do with a pass over later to see what state the various separate encryption tests are in)

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Encryption/ZipAESEncryptionStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Encryption/ZipAESEncryptionStream.cs
@@ -1,0 +1,140 @@
+using System;
+using System.IO;
+using System.Security.Cryptography;
+
+namespace ICSharpCode.SharpZipLib.Encryption
+{
+	/// <summary>
+	/// Encrypts AES ZIP entries.
+	/// </summary>
+	/// <remarks>
+	/// Based on information from http://www.winzip.com/aes_info.htm
+	/// and http://www.gladman.me.uk/cryptography_technology/fileencrypt/
+	/// </remarks>
+	internal class ZipAESEncryptionStream : Stream
+	{
+		// The transform to use for encryption.
+		private ZipAESTransform transform;
+
+		// The output stream to write the encrypted data to.
+		private readonly Stream outputStream;
+
+		// Static to help ensure that multiple files within a zip will get different random salt
+		private static readonly RandomNumberGenerator _aesRnd = RandomNumberGenerator.Create();
+
+		/// <summary>
+		/// Constructor
+		/// </summary>
+		/// <param name="stream">The stream on which to perform the cryptographic transformation.</param>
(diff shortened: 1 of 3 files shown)


Numpsy:
not sure what happend to the CI build there, it says

```
Exception calling "UploadFile" with "2" argument(s): "Unable to connect to the remote server"
```

which sounds like a server side error

Numpsy:
With regards to the AddExtraDataAES bit - is there any value in adapting that into the TaggedData machinery (ExtendedUnixData/NTTaggedData etc) rather than having the structure inline at the points where it's read/written?

piksel:


```diff

Complexity increasing per file
==============================
- src/ICSharpCode.SharpZipLib/Encryption/ZipAESEncryptionStream.cs  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745481&bid=18799040)

Numpsy:
Some things to be decided:

1) The API to use for enabling AES encryption.
I was working #380 towards adding a common enum for both apis, so possibly a property of type ```ZipEncryptionMethod``` would be sufficient?

2) The logix for when to 'apply' encryption to the entries (does it set the AESKeySize properties on new entries in the Add functions, as soon as the entries are added, or when the updates are commited, when it knows both the current setting and current password)

It also possibly needs checking for how it handles (or not) directory entries that have AESKeySize set - it doesn't do the encryption machinery for folders like ZipOutputStream does, so might need to review things like EncryptionMethodForHeader being used for directories.

piksel:
1. Yeah, I think that's ideal

2. I would presume that setting it when committing the updates is less error-prone. Do you see any down sides to it?

Numpsy:
I can't think of any, just need to document that entries are encrypted based on the value of the setting at that point (in case anyone gets the idea that you can add multiple entries with different and/or some/no encryption in one go).

The ZipCrypto case already decides what to do based on if the password is set when the updates are run, so it's consistent with that as well.

PR #445, opened 2020-04-12
Make InvalidHeaderException serializable

Numpsy:
I was looking at some code analyzer results of the build, and one of the things it suggested was that InvalidHeaderException should be [Serializable], given that it derives from a serializable type (TarException).

I don't know why this wasn't done along with the others, aside from having been overlooked?


piksel:
Yeah, probably overlooked, no reason for any `Exception` to not be marked as `Serializable`

PR #449, opened 2020-04-16
[WIP] Add FxCop to the build

Numpsy:
Add FxCop to the build to see what issues it finds.

Marked as WIP because it creates over a thousand warnings in the build.

Some of them look worth looking at (e.g. warnings about not disposing of things before they go out of scope), but a lot of them are things like passing string literals to exceptions rather than using resources, which are possibly not really useful. (so could be turned off)

(Note: #445 is something that was found by the analyzer)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -37,3 +37,9 @@ end_of_line = crlf
 [*.yml]
 indent_style = space
 indent_size = 2
+
+# C# source
+[*.{cs}]
+
+# CA1303: Do not pass literals as localized parameters
+dotnet_diagnostic.CA1303.severity = none
--- a/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
+++ b/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -31,6 +31,10 @@ Please see https://github.com/icsharpcode/SharpZipLib/wiki/Release-1.3.2 for mor
   </PropertyGroup>
 
 	<ItemGroup>
+		<PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.2">
+		  <PrivateAssets>all</PrivateAssets>
+		  <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+		</PackageReference>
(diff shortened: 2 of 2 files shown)


piksel:
Good initiative. I basically did this and turned it off again because of the spamming, but we should set up a ignore list, including CA1303 (until we get localization).

Numpsy:
Is it ok to use the editorconfig approach to control the analyzer settings?
(It seems to be the MS recommended way now in VS 2019, but I can't say offhand if that works in 2017)

Numpsy:
I take the way that the number of warnings in the CI build didn't go down after updating editorconfig to mean the answer to my previous question is 'no' :-(

In any case, I sent #450 to fix some of the produced warnings.

Numpsy:
@piksel I left this using the 2.9 version of FxCop as the release notes for the v3 version say it requires VS 2019 and I thought that may not be wanted, but maybe that's not an issue now the tests are targetted at Core 3.1 and need that anyway?

piksel:
Yeah, I don't see any problems with requiring VS 2019 for FxCop. If a need for this to change emerges then we could just disable or change this later, but for now I don't see any reason to use anything earlier than 2019.

Numpsy:
hmm, I don't know how this change could cause that one test failure.

piksel:
No, that's probably due to something unexpected in the CI. Especially since it worked on the same platform, different config, and same config, different platform(s).

piksel:
I'm thinking about how to best get this merged. The best thing to do is probably to just create a bulk commit/PR with most of the fixes done. This will render other PRs in need of rebasing, but at least it would be more of an one-time event.

Numpsy:
I think a pretty large proportion of the remaining warnings were about checking public params for being null, and at the time I might have thought that #501 would deal with some of those and not looked in much detail.
I'll have a look, and see what proportion are 'non-functionality' ones.

Numpsy:
Theres also the question at this point on if it should use NetAnalyzers instead of FxCop (I can see about a PR that fixes many warnings at once in either case though))

SourceproStudio:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #451, opened 2020-04-18
Update and minimize sample project package dependencies

Numpsy:
I was browsing around the project insights github page and noticed that the package dependencies listed on the [dependancy graph](https://github.com/icsharpcode/SharpZipLib/network/dependencies) page looked rather seriously excessive, so I thought a spring clean might be in order.

(Looking at the history, it looks like the extra deps come are as a result of pointing the .net 4.6.1 sample projects at a build of the lib that only targeted .NET Standard 1.3, so shouldn't matter not the lib targets .NET 4.5 again)


Numpsy:
Updated to change all the C# sample projects, and to change a couple of uses of ```ZipConstants.DefaultCodePage``` to ```ZipStrings.CodePage``` to fix warnings about using obsolete functions.

PR #452, opened 2020-04-21
In ZipFile.GetOutputStream, ensure that any created crypto streams are disposed

Numpsy:
When looking at #443 I noticed that the encryption streams created in ZipFile.GetOutputStream didn't seem to be getting disposed.

In order to ensure that they are, I think we need to

1) In the CompressionMethod.Deflated case, make the DeflaterOutputStream own the encryption stream, so that they get disposed together.

2) In the CompressionMethod.Stored case, I think it should be safe to return the encryption stream directly in this case rather than creating an intermediate UncompressedStream? (The stream is only used internally so there should be minimum scope for unexpected uses of it).


Numpsy:
(443 contains a variant of this change, this is to break it out into it's own piece as it seems useful to have separately)

PR #458, opened 2020-05-03
Dispose of entry streams returned by ZipFile.GetInputStream

Numpsy:
Another offshoot of #449 - When extracting a zip with FastZip, dispose of the streams returned from ZipFile.GetInputStream().


piksel:
This is not *actually* needed, but Stream is `IDisposable`, so I guess it's better to comply :3

Numpsy:
I thought it best to be 'nice' about tidying it up, as if it's doing AES decryption for the entry, the stream might contain crypto classes that are disposable as well (though I think there are some FxCop warnings about things in there not being disposed)

Issue #454, opened 2020-04-22
What is the encryption that is used by FastZip?


zivsha:
Can anyone tell me what is the default encryption used when setting a password for a `FastZip`?
I'm using v1.1.0.
Is there a way to specify which encryption to use when using FastZip?
Or at least which AES to use?

Numpsy:
The default would be ZipCrypto.

I was working on adding a direct means of setting the encryption method to use in #380, but that is still a work in progress, and there is no direct way to configure the algorithm in the current version.

Numpsy:
(I say 'direct' because reading this has just given me the idea that you might be able to do it by setting the EntryFactory property of the FastZip instance to a factory that enables AES on file entries, i'll have a think about at that)

Numpsy:
hmm, in the short term you could try defining something like

```
class AesEntryFactory : IEntryFactory
{
    readonly ZipEntryFactory factory = new ZipEntryFactory();

    INameTransform IEntryFactory.NameTransform
    {
        get => factory.NameTransform;
        set => factory.NameTransform = value; 
    }

    ZipEntry IEntryFactory.MakeDirectoryEntry(string directoryName) => factory.MakeDirectoryEntry(directoryName);

    ZipEntry IEntryFactory.MakeDirectoryEntry(string directoryName, bool useFileSystem) => factory.MakeDirectoryEntry(directoryName, useFileSystem);

    ZipEntry IEntryFactory.MakeFileEntry(string fileName)
    {
        var entry = factory.MakeFileEntry(fileName);
        entry.AESKeySize = 256;

        return entry;
    }

    ZipEntry IEntryFactory.MakeFileEntry(string fileName, bool useFileSystem)
    {
        var entry = factory.MakeFileEntry(fileName, useFileSystem);
        entry.AESKeySize = 256;

        return entry;
    }

    ZipEntry IEntryFactory.MakeFileEntry(string fileName, string entryName, bool useFileSystem)
    {
        var entry = factory.MakeFileEntry(fileName, entryName, useFileSystem);
        entry.AESKeySize = 256;

        return entry;
    }
}
```

and then pointing FastZip.EntryFactory at that (to set the encryption to AES 256, set to 128 by setting AESKeySize to 128 instead)

zivsha:
Isn't it enough to just set the `AESKeySize` for each `ZipEntry` when the `FastZip` is processing folders and files?

Numpsy:
[...]

PR #460, opened 2020-05-11
In ZipOutputStream.PutNextEntry, account for AES overhead when calculating compressed entry size

Numpsy:
Found when looking at #454:

The code at:

https://github.com/icsharpcode/SharpZipLib/blob/f36ca373206340a25b1d3bb226acc7b679128a7c/src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs#L330

for writing entries when the header information is known tries to take account of the crypto header overhead, but only does so for ZipCrypto encrypted entries - it uses the wrong values for AES entries.

Also, the code for building the Zip64 extra data at:

https://github.com/icsharpcode/SharpZipLib/blob/f36ca373206340a25b1d3bb226acc7b679128a7c/src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs#L375

doesn't take account of that overhead at all, which seems to also produce incorrect archives.

~~The~~

```
// TODO: Refactor header writing.  Its done in several places.
```
~~becomes a bit more true with the simple change, as it duplicates a bit more logic that could perhaps do with being shared.~~

PR adds a set of test cases for different encryption/compression methds, with and without zip64, but only for 0 byte entries - something with populated entries whose size and crc are known upfront might be useful seperately.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
@@ -655,7 +655,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 
 					if ((versionToExtract == 0) && IsCrypted)
 					{
-						trueCompressedSize += ZipConstants.CryptoHeaderSize;
+						trueCompressedSize += (ulong)this.EncryptionOverheadSize;
 					}
 
 					// TODO: A better estimation of the true limit based on compression overhead should be used
@@ -1010,6 +1010,26 @@ namespace ICSharpCode.SharpZipLib.Zip
 				// Variable		Encrypted file data
 				//    10		Authentication code
 				return 12 + AESSaltLen;
+			}
+		}
+
+		/// <summary>
+		/// Number of extra bytes required to hold the encryption header fields.
+		/// </summary>
+		internal int EncryptionOverheadSize
+		{
+			get
+			{
+				// Entry is not encrypted - no overhead
+				if (!this.IsCrypted)
+					return 0;
+
+				// Entry is encrypted using ZipCrypto
+				if (_aesEncryptionStrength == 0)
+					return ZipConstants.CryptoHeaderSize;
+
+				// Entry is encrypted using AES
+				return this.AESOverheadSize;
 			}
 		}
 
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs
@@ -337,7 +337,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 				}
 				else
 				{
-					WriteLeInt(entry.IsCrypted ? (int)entry.CompressedSize + ZipConstants.CryptoHeaderSize : (int)entry.CompressedSize);
+					WriteLeInt((int)entry.CompressedSize + entry.EncryptionOverheadSize);
 					WriteLeInt((int)entry.Size);
 				}
 			}
@@ -382,7 +382,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 				if (headerInfoAvailable)
 				{
 					ed.AddLeLong(entry.Size);
-					ed.AddLeLong(entry.CompressedSize);
+					ed.AddLeLong(entry.CompressedSize + entry.EncryptionOverheadSize);
 				}
 				else
 				{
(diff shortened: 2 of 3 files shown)


Numpsy:
Updated to move some of the duplicated overhead size calculation logic to an 'EncryptionOverheadSize' property on ZipEntry, and also changed ZipEntry.LocalHeaderRequiresZip64 to use that instead of always using the ZipCrypto constant.

Numpsy:
A small observation about the length calculation logic ZipOutputStream.CloseEntry that I noticed when looking at this:

When calculating the compressed size of the entry, it does

```
if (curEntry.IsCrypted)
{
	if (curEntry.AESKeySize > 0)
	{
		curEntry.CompressedSize += curEntry.AESOverheadSize;
	}
	else
	{
		curEntry.CompressedSize += ZipConstants.CryptoHeaderSize;
	}
}
```

but then after that it does

```
if (curEntry.LocalHeaderRequiresZip64)
```

and inside ```LocalHeaderRequiresZip64```, it does

```
if ((versionToExtract == 0) && IsCrypted)
{
    trueCompressedSize += ZipConstants.CryptoHeaderSize;
}
```

So it looks like it counts the crypto overhead twice.
Not sure if there's any scope for an edge case there where that makes it use the wrong values if those few exctra bytes push it over the non-64 length limit?

src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
```diff
@@ -1013,6 +1013,26 @@ internal int AESOverheadSize
 			}
 		}
 
+		/// <summary>
+		/// Number of extra bytes required to hold the encryptio header fields.
```
piksel, line 1017:
```suggestion
		/// Number of extra bytes required to hold the encryption header fields
```

src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs
```diff
@@ -327,7 +327,14 @@ public void PutNextEntry(ZipEntry entry)
...
 				{
-					WriteLeInt(entry.IsCrypted ? (int)entry.CompressedSize + ZipConstants.CryptoHeaderSize : (int)entry.CompressedSize);
+					int entryCompressedSize = (int)entry.CompressedSize;
+
+					if (entry.IsCrypted)
+					{
+						entryCompressedSize += entry.EncryptionOverheadSize;
+					}
+
+					WriteLeInt(entryCompressedSize);
```
piksel, line 337: Why check IsCrypted twice?
```suggestion
					WriteLeInt((int)entry.CompressedSize + entry.EncryptionOverheadSize);
```

src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs
```diff
@@ -372,7 +379,15 @@ public void PutNextEntry(ZipEntry entry)
...
-					ed.AddLeLong(entry.CompressedSize);
+
+					long entryCompressedSize = entry.CompressedSize;
+
+					if (entry.IsCrypted)
+					{
+						entryCompressedSize += entry.EncryptionOverheadSize;
+					}
+
+					ed.AddLeLong(entryCompressedSize);
```
piksel, line 390: Could be simplified (since `IsCrypted` is already checked in `EncryptionOverheadSize`)
```suggestion
					ed.AddLeLong(entry.CompressedSize + entry.EncryptionOverheadSize);
```

Numpsy:
ok, too much copy/pasting, will have a look over the weekend.

Numpsy:
updated

src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
```diff
@@ -1013,6 +1013,26 @@ internal int AESOverheadSize
...
+		internal int EncryptionOverheadSize
+		{
+			get
+			{
+				// Entry is not encrypted - no overhead
+				if (!this.IsCrypted)
+					return 0;
+
+				// Entry is encrypted using ZipCrypto
+				if (_aesEncryptionStrength == 0)
```
piksel, line 1028: Codacy found an issue: [Add curly braces around the nested statement(s) in this 'if' block.](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745474)

piksel:


```diff

Issues
======
- Added 2
           

Complexity decreasing per file
==============================
+ src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs  -1
         

Clones removed
==============
+ src/ICSharpCode.SharpZipLib/Zip/ZipOutputStream.cs  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745474&bid=18799033)

src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
```diff
@@ -1013,6 +1013,26 @@ internal int AESOverheadSize
...
 
+		/// <summary>
+		/// Number of extra bytes required to hold the encryption header fields.
+		/// </summary>
+		internal int EncryptionOverheadSize
+		{
+			get
+			{
+				// Entry is not encrypted - no overhead
+				if (!this.IsCrypted)
```
piksel, line 1024: Codacy found an issue: [Add curly braces around the nested statement(s) in this 'if' block.](https://app.codacy.com/manual/piksel/SharpZipLib/pullRequest?prid=5745474)

Issue #40, opened 2013-12-29
Create/Extract NTFS Extra Data field for Timestamp etc


dgrunwald:
[SD-1800](http://bugtracker.sharpdevelop.net/issue/ViewIssue.aspx?id=1800), originally created on 12/20/2010 01:58:10 by David Pierson

We should have the ability to create the NTFS Extra Data field (0x000a)
via an option in FastZip, and an easy way to create it in ZipFile and
ZipOutputStream. 

 This would hold the NTFS Last Modified timestamp, avoiding the problem
of the 2 second granularity of the standard entry timestamp. 

 Relevant forum threads:

[http://community.sharpdevelop.net/forums/t/12411.aspx](http://community.sharpdevelop.net/forums/t/12411.aspx)

[http://community.sharpdevelop.net/forums/t/14178.aspx](http://community.sharpdevelop.net/forums/t/14178.aspx)

Numpsy:
I see that there is already an 'NTTaggedData' class in the extra data handling that appears to handle the NTFS extra data block, but the only place it's referenced from the library is the ifdefed out block at:

https://github.com/icsharpcode/SharpZipLib/blob/fed3bd219f8bd2bac5287b6217b9c4c384bed35f/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs#L1103

There is a comment there about disabling it by default to match InfoZip, but i'm not sure what the logic should be at this point - is there a reason to not use the data if present in the zip? (or maybe restrict it to extraction on Windows hosts?)

piksel:
No idea. I have had it on my agenda to look through the extra data fields...

bastianeicher:
@Numpsy I added that `#if RESPECT_NT_TIMESTAMP` a long time ago for an admittedly rather specific use case:

Zero Intall is a cross-platform package manager. The Linux version uses CLIs like `tar` and `unzip` (InfoZIP) to extract archives. [Zero Install for Windows](github.com/0install/0install-dotnet) uses SharpZipLib. After extracting archives (ZIP, TAR, etc.) both versions verify the files using checksums which include the timestamps. Therefore I needed SharpZipLib to produce identical timetamps to InfoZIP.

Perhaps a better, although still rather clumsy, alternative to the `#if` block might be a `public static bool` config toggle.

Numpsy:
[...]

PR #463, opened 2020-05-20
Improve support for Unix timestamps in ZIP archives

bastianeicher:

Changes:

- Store `ZipEntry.DateTime` in dedicated backing field.  
  This allows reading values with a higher resolution than DOS time (2 second accuracy).
- Fix getting unix modification time in ZIP files.  
  I was the contributor that originally added that if-check, but I unfortunately was mistaken. InfoZIP actually does respect a file's modification time, even if the access time and/or creation time are not set.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
@@ -1103,12 +1103,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 
 			// Check for Unix timestamp
 			ExtendedUnixData unixData = extraData.GetData<ExtendedUnixData>();
-			if (unixData != null &&
-				// Only apply modification time, but require all other values to be present
-				// This is done to match InfoZIP's behaviour
-				((unixData.Include & ExtendedUnixData.Flags.ModificationTime) != 0) &&
-				((unixData.Include & ExtendedUnixData.Flags.AccessTime) != 0) &&
-				((unixData.Include & ExtendedUnixData.Flags.CreateTime) != 0))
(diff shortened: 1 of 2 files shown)


Numpsy:
Seems like a reasonable simplification (avoids doing all those calculations again on each DataTime property get), and possible helpful for #40 as well (related to the other ifdefed out data time code @ https://github.com/icsharpcode/SharpZipLib/blob/d5d1a52a4f95fbba156e92a39116b1bb5bed8862/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs#L1099)

It does appear to have caused a pair of unit test failures though

```
ICSharpCode.SharpZipLib.Tests.Zip.ZipEntryHandling.DateAndTime
ICSharpCode.SharpZipLib.Tests.Zip.ZipEntryHandling.Copying
```

(the test failures don't cause the 'checks' to fail alas).

Looking at the code, I was wondering if there is scope for the DateTime and DosTime members to get out of sync (where setting DateTime updates DosTime, but not vice-versa)?

bastianeicher:
@Numpsy: Thanks for the fast feedback. :)

I've added an additional commit to the PR that removes the backing field for `ZipEntry.DosTime`. The getters and setters instead convert to and from `ZipEntry.DateTime` to prevent the two values from becoming potentially inconsistent.

This also fixes the failing unit tests. Sorry I missed them in the original PR.

bastianeicher:
Anything else I can do to help this PR get merged?

piksel:
Sorry, I have this on my backlog to go through. I hope to get to it soon!

piksel:
@bastianeicher Could you check my fix in #502? I changed the `DateTime.Kind` to `Unspecified` instead of `UTC` to prevent manually set dates from being shifted after "Serializing" them in an archive.

PR #465, opened 2020-05-22
Fixed bug in ZipAESStream.ReadBufferedData.

Helios-vmg:
copyCount was being computed but not passed to Array.Copy(), which caused an
exception to be thrown when enough bytes were read from an encrypted stream.


Numpsy:
My bad, I made some changes in that area to fix some issues with buffer sizes and apparently both missed that case and left a hole in the unit tests :-( 

I've pushed a unit test change that reproes the issue as #466.

PR #467, opened 2020-05-23
Allow seeking a PartialInputStream to the very end.

Helios-vmg:
This is useful when passing the object to C code, as applications often seek() to the end and tell() to find out the length of a stream.


Numpsy:
Hmm, does the current code throw if you do Seek(0, SeekOrigin.End)? (Doesn't seem right if so) 

(/Wonders if it should really be throwing at all when you try to seek past the end at all - reading the [MS Docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.seek?view=netframework-4.7.2) to remind myself, it says ```Seeking to any location beyond the length of the stream is supported```, so perhaps it shouldn't)

Helios-vmg:
Yes, it does throw on Seek(0, End). I think, particularly for an input stream, throwing when seeking beyond the end is useful to detect errors early.

Issue #471, opened 2020-06-09
Can't open an AES-encrypted Zip archive with password set to empty string. Is this archive nonstandard?


michaelrgibbs:
### Steps to reproduce
1. Create a zip file with AES-256 encryption and an empty password. I have some created in the past with DotNetZip. I don't think they can be created with SharpZipLib.
2. Create a ZipFile object from the zip file and set the Password field to "".
3. Try to extract.

### Expected behavior
The zip file is opened and extracted.

### Actual behavior
The code acts as if no password has been set and throws exception "No password available for AES encrypted stream".

I see the setter code doesn't differentiate between null and empty string: https://github.com/icsharpcode/SharpZipLib/blob/master/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L360

7zip and DotNetZip allow these files to be extracted with an empty password, but I can't work out if they are just being more lenient than they need to be.

### Version of SharpZipLib
1.2.0

Numpsy:
Not sure why you'd want to be able to create an archive like that, but the WinZip AES spec doesn't seem to say you can't, so maybe something to handle in a read only fashion?

As fas as SharpZipLib goes:

The AES decryption machinery itself seems to handle it if you can get the empty string password into it, but the ZipCrypto machinery does require a value (that's the ```GenerateKeys``` call in the setter you link to).

Passing an empty value into GenerateKeys causes it to throw (hence the way the null/empty check works I suppose), but that wouldn't matter for AES decryption as that doesn't use that key, it just uses the password string directly.

Related observation: Doesn't the way that setter works mean that if you set the password to some string value, and then set it to null, that the ```rawPassword_``` string will get left at the previous value?

The AES encryption setup @ https://github.com/icsharpcode/SharpZipLib/blob/195b4e2275f9d4d486b44add324d574c851db2a3/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L3617 also checks if the 'key' has been setup but then sets up the decryption using the password string instead of that, which is also maybe not correct...

Maybe the Password setter should always set ```rawPassword_``` to the passed value, and ```CreateAndInitDecryptionStream``` for AES entries should just check for the string being non-null ?

michaelrgibbs:
I think it's a bug that creates them in the first place, but I want to support archives created in the past when the bug was present.

piksel:
[...]

PR #472, opened 2020-06-14
Allow ZipFile to accept empty strings as passwords when decrypting AES entries

Numpsy:
refs #471 

Two changes:

1) In the current setter for ```ZipFile.Password```, it only updates the ```rawPassword_``` member if the new value isn't null or empty.
That seems wrong to me in general, as it means that setting the password to a value and then setting it to null would leave the member set to the old value?
This PR changes it to always just save the new value in ```rawPassword_```.

2) In CreateAndInitDecryptionStream, throw if the string password is null, but proceed to try to use it if the string is empty.

The current code uses ```HasKeys``` to check that it has the keys, but then uses the password instead, which doesn't sound quite right? (though looking at the way ```KeysRequired```/```OnKeysRequired``` I'm not sure if it works to get the password (rather than the keys') for decrypting AES entries?

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -367,9 +367,10 @@ namespace ICSharpCode.SharpZipLib.Zip
 				}
 				else
 				{
-					rawPassword_ = value;
 					key = PkzipClassic.GenerateKeys(ZipStrings.ConvertToArray(value));
 				}
+
+				rawPassword_ = value;
 			}
 		}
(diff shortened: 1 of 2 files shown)


Numpsy:
hmmm, that's an unfortunate test failure in the CI build

```
  Expected: "..., consectetur adipiscing elit.\nFusce bibendum diam ac nunc..."
  But was:  "..., consectetur adipiscing elit.\r\nFusce bibendum diam ac nun..."
```

Not sure why it's \n on the CI system, but maybe that means the test would fail on a non-windows host and needs fixing anyway.

Issue #405, opened 2019-12-30
FastZip does not allow Extended Path Filter to be specified


atcobb01:
Version of SharpZipLib: 1.2.0

trying to use Extended Path Filter, but don't see any examples.  FastZip uses strings for Name and Directory filters, but I need file time as a criteria.  I don't see how to use the ExtendedPathFilter class when creating a zip file using FastZip or any other method.  Can you help?

piksel:
No, there does not seem to be any way to provide custom filters for `FastZip`. I'm not sure about what the purpose of `ExtendedPathFilter` is. It looks like it's meant to be used in `FastZip`, perhaps with an overloaded constructor. That should be a pretty small PR. 

Otherwise, your use case could be achieved fairly easily with `ZipFile` like this:
```cs
using (var zf = new ZipFile(zipFileName)) {
    zf.BeginUpdate();
    var yesterday = DateTime.UtcNow.AddDays(-1);
    var files = Directory.GetFiles(sourceDir)
        .Where(f => new FileInfo(f).LastWriteTimeUtc > yesterday);
        
    foreach(var file in files){
        // Provide an entry name as a second parameter 
        // to avoid using the full path to the file
        var entryName = Path.GetFileName(file);
        zf.Add(file, entryName);
    }
    zf.CommitUpdate();
}
```

Numpsy:
Maybe an extra version of ```CreateZip``` that takes the file/directory filters as ```IScanFilter``` rather than strings?

It looks like ```FileSystemScanner``` already has a constructor that can take a pair of IScanFilters on top of the one that uses strings (currently used by FastZip), so maybe that's a straight forward way of passing in alternate filter implementations?

Numpsy:
This one should be fixed by #482

piksel:
This should be fixed by #482.

PR #482, opened 2020-07-14
Add variants of FastZip.CreateZip that take filters as IScanFilter instead of strings

Numpsy:
refs #405: Add simple extra versions of FastZip.CreateZip that take file/directory filters as instances of IScanFilter rather than strings.

No unit tests yet (There don't seem to be any tests for ExtendedPathFilter either?)

A possible small api compatibility issue: if you have a call like

```csharp
zippy.CreateZip(tempZip.Filename, tempDir.Fullpath, true, null, null);
```
(as in an existing unit test I had to change), the two null literals for the filter parameters are ambiguous between the existing ```string``` overload and the new ```IScanFilter``` overload, which could possibly need a change to avoid?

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
@@ -376,6 +376,49 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// <param name="leaveOpen">true to leave <paramref name="outputStream"/> open after the zip has been created, false to dispose it.</param>
 		public void CreateZip(Stream outputStream, string sourceDirectory, bool recurse, string fileFilter, string directoryFilter, bool leaveOpen)
 		{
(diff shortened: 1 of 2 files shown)


Numpsy:
actually, I suppose there would also be a (sort of) conflict with #455 where that adds an extra overload with a leaveOpen param to the string-filter variants, and this would leave it without such an overload for the IScanFilter variants (hurrah for overload explision :-( )

piksel:
Yeah, but both variants are really nice to have. The `IScanFilter` overload could just take the leaveOpen with a default value instead, so it shouldn't be that bad.

Numpsy:
I'll get this updated tomorrow

Numpsy:
Rebased on top of the other changes, and changed the version of CreateZip that takes an output stream and the ScanFilters to have a ```leaveOpen``` parameter that defaults to false.

PR #497, opened 2020-08-05
Give ZipOutputStream an INameTransform property, and use it to transform the names of newly added entries

Numpsy:
Ref: the discussions about name transforming/cleaning when creating zips:

An experiment into giving ZipOutputStream an INameTransform to (optionally) use to transform file paths, to see what it might look like.
A bit more heavyweight that just having a clean function to call, but more flexible (if that's useful), and fits in more with ZipFile and FastZip?

Defaults to a minimized transform that just strips path roots and puts slashes in the appropriate format (copied from the existing versions and trimmed a bit basically).

ZipEntry.Name is read-only after creation, so it ends up transforming the name twice when doing the local/central headers, rather than just doing it once.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs
@@ -383,6 +383,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 			{
 				outputStream_.SetLevel((int)CompressionLevel);
 				outputStream_.IsStreamOwner = !leaveOpen;
+				outputStream_.NameTransform = null; // all required transforms handled by us
 
 				if (password_ != null)
 				{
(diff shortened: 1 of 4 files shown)


Numpsy:
(On a related note, the existing TransformFile function could perhaps use the string trim functions instead of manual loops for trimming leading/trailing separators?)

Numpsy:
refs #495

piksel:
Yeah, I really like this solution, as it gives a good default with a transparent implementation (`PathTransform`) that still leaves the consumer with the flexibility to customize the behaviour if they so choose.

Numpsy:
ZipEntry doesn't expose a means to set the name after construction, so ZipOutputStream couldn't just mutate it in PutNextEntry.

Could maybe give ZipEntry.Name an internal setter to use to update it?

piksel:
Although, if we allow the `null` filter, then we might as well allow changing the Name property (for non-commited ZipEntries of course, since changing it for one that is already written to a zip file won't have the intended effect and would break the lookup).

Numpsy:
Updated to rebase on top of the other changes, and to try the alternate approach of an internal property setter for ZipEntry.Name and only doing the transform once in ZipOutputStream.

piksel:
Yeah, this is probably the easiest to do right now. Allowing rename on uncommited entires introduces a lot of complexity for no real gain. LGTM!

PR #501, opened 2020-08-06
Enable explicit null checking of reference types

piksel:
Adds explicit null checking for reference types using C# 8.

Running [InferNull](https://github.com/icsharpcode/NullabilityInference) on the main source project made it go down from 661 warnings to 214.
After some cleaning and refactoring, the project now happily compiles without any warnings and with almost complete null-checking (some places were just explicitly ignored since they would have a much greater performance impact).


Numpsy:
With regards to the ```EmptyRefs``` thing:

When looking at the FxCop/analyzer warning stuff I saw multiple suggestions to replace allocations of [0] length arrays with Array.Empty<>.
Didn't try to do anything with it yet due to the lack of Array.Empty in NET45, but it did cross my mind to see if it was worth/possible to add an internal definition of Array.Empty in that version, and then just use Array.Empty  in the rest of the code, which wouldn't have to be different on different targets then?

piksel:
> [...] add an internal definition of `Array.Empty` in that version, and then just use `Array.Empty` in the rest of the code, which wouldn't have to be different on different targets then?

Since `Array` is defined in `System` I just disregarded that idea, but now that you mention it, it could be added as an extension method and then keeping the actual instances in the static class.

Numpsy:
I thought i'd seen some other project do that, but maybe i'm misremembering, though the implementation is apparently only something like
```csharp
internal static class EmptyArray<T>
{
	public static readonly T[] Value = new T[0];
}
```
anyway.

src/ICSharpCode.SharpZipLib/GZip/GZip.cs
```diff
@@ -18,7 +18,7 @@ public static class GZip
 		/// <param name="outStream">The output stream to receive the decompressed data.</param>
 		/// <param name="isStreamOwner">Both streams are closed on completion if true.</param>
 		/// <exception cref="ArgumentNullException">Input or output stream is null</exception>
-		public static void Decompress(Stream inStream, Stream outStream, bool isStreamOwner)
+		public static void Decompress(Stream? inStream, Stream? outStream, bool isStreamOwner)
```
Numpsy, line 21: should these streams be nullable if the function throws on null?

piksel: Not sure, I don't think we can check for null if we say that they're non-nullable. I have to read up on how to interface with consumer code with and without Nullable enabled.

piksel:
> [...] though the implementation is apparently only something like  

Yeah, that is both the implementation in [mscorlib](https://referencesource.microsoft.com/#mscorlib/system/array.cs,3080) and the way I polyfilled it now. I haven't pushed those changes yet though, since I'm not happy with the naming yet.

src/ICSharpCode.SharpZipLib/GZip/GZip.cs
Numpsy: Callers without Nullable enabled will just behave as they do before (i.e treat it as though it can be null), and you can still check for null and throw for those cases (or in general really, given that the checks don't actually *prevent* it from being null even if marked as non-nullable).

FxCop still gives warnings that non-nullable parameters to public functions should be checked before use as well.

piksel: Yep you're right. There are some instances where you branch on null that the compiler thinks it should be nullable.

Numpsy:
Would it be worth breaking the Array.Empty bits out into a seperate change? (it could pass as a trivial optimization, and shou;dn't rely on the other nullability changes?)

piksel:
Closing this as it would have to be redone now anyway.

PR #503, opened 2020-08-08
Consider AES overhead when testing encrypted folder entries

Numpsy:
refs my last comment in #317 - TestArchive doesn't seem to consider AES encryption when testing encrypted directory entries (it always uses ```ZipConstants.CryptoHeaderSize```).

This is just a simple change to use the newly added ```EncryptionOverheadSize ``` instead.

I'm not sure what the '+ 2' is allowing for? (empty compressed data overhead? I'm not sure what the comment in the non-encrypted case about not knowing the true size refers to, when directories don't really have a size?)


Numpsy:
(tests presently fail on master because of #353 - might need to finish that one first for testing purposes, or merge the PRs together)

Issue #507, opened 2020-08-12
Attempting to use AES encryption with no password throws unhelpful  exception


piksel:
> I tried turning off the password in one of the zip output stream unit tests and got
>
> ```
> Message: 
>   System.ArgumentNullException : Buffer cannot be null.
>   Parameter name: buffer
> Stack Trace: 
>   MemoryStream.Write(Byte[] buffer, Int32 offset, Int32 count)
>   ZipOutputStream.CloseEntry() line 534
>   ZipOutputStream.Finish() line 772
>   DeflaterOutputStream.Dispose(Boolean disposing) line 404
>   Stream.Close()
>   ZipEncryptionHandling.WriteEncryptedZipToStream(Stream stream, String password, Int32 keySize, CompressionMethod compressionMethod) line 497
>   ZipEncryptionHandling.CreateZipWithEncryptedEntries(String password, Int32 keySize, CompressionMethod compressionMethod) line 540
> ```
>
> Because ```AESAuthCode``` is null.
> 
> Looks to be because some of the logic goes off entry.AESKeySize being set to a non-zero value (deciding whether to write the > AES extra data etc), but them some of it goes off Entry.IsCrypted being true, which is only done automatically if a password has already been specified when the entry is added, so some of the other crypto bits get skipped.
> 
> Gets itself in a bit of a mess basically. 
> Possibly ```PutNextEntry``` should throw if you try to add an entry with AESKeySize set, but no password set?

_Originally posted by @Numpsy in https://github.com/icsharpcode/SharpZipLib/discussions/506#discussioncomment-49971_

piksel:
Fixed by #509

PR #509, opened 2020-08-17
Make PutNextEntry throw if the entry is AES and no password has been set. 

Numpsy:
refs #507 - possible simple way to avoid the issue, by making PutNextEntry throw if the new entry has an AESKeySize set and the stream password hasn't been set.

I made it an InvalidOperationException rather than an ArgumentException with the idea that the operation is invalid when the stream has no password, 


piksel (approved):
Yeah, the argument isn't actually the problem, but rather the state of the ZipOutputStream (even though you could argue that the AESKeySize of the argument is the issue, but I think that's less helpful).

PR #511, opened 2020-08-17
Move the 7zip helper functions out of ZipEncryptionHandling into their own class, for easier use elsewhere

Numpsy:
When I was looking at the ZipFile/BZip2 changes I was thinking that it might be useful to verify some of those created archives with 7zip in the way it's currently done for testing encrypted archives, so - I thought it might be useful to break the 7zip helper functions out into their own class rather than being inside the ```ZipEncryptionHandling``` test class as a start towards that.


piksel (approved):
Yeah, agreed.

PR #517, opened 2020-10-03
Throw exception on Store+Descriptor entries

### LINKED ISSUE #15: ZipInputStream class doesn't process correctly no compression entries with flag Bit 3 set
The PKZIP documentation [here](http://www.pkware.com/documents/casestudies/APPNOTE.TXT) states for general purpose flag Bit 3:

"If this bit is set, the fields crc-32, compressed size and uncompressed size are set to zero in the local header.  The correct values are put in the data descriptor immediately following the compressed data.  (Note: PKZIP version 2.04g for DOS only recognizes this bit for method 8 compression, newer versions of PKZIP recognize this bit for any compression method.)".

---

The CRC, compressed and uncompressed size will be 0 and the extraction fails.

piksel:
Due to the way that ZipInputStream is designed, reading entries that are using `Store` compression method and `Descriptor` flag is not possible. This change will throw a `StreamUnsupportedException` instead of failing on incorrect CRC.

Fixes #15 (as much as a fix as is possible for this).

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
@@ -722,14 +722,13 @@ namespace ICSharpCode.SharpZipLib.Zip
 		}
 
 		/// <summary>
-		/// Gets/Sets the compression method. Only <see cref="CompressionMethod.Deflated">Deflated</see>
-		/// and <see cref="CompressionMethod.Stored">Stored</see> are supported.
+		/// Gets/Sets the compression method.
 		/// </summary>
+		/// <remarks>Throws exception when set if the method is not valid as per <see cref="IsCompressionMethodSupported()"/></remarks>
+		/// <exception cref="NotSupportedException"/>
 		/// <returns>
 		/// The compression method for this entry
 		/// </returns>
-		/// 
-		/// 
 		public CompressionMethod CompressionMethod
 		{
 			get => method;
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
@@ -263,13 +263,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// <summary>
 		/// Get a value indicating whether the entry has a CRC value available.
 		/// </summary>
-		public bool HasCrc
-		{
-			get
-			{
-				return (known & Known.Crc) != 0;
-			}
-		}
+		public bool HasCrc => (known & Known.Crc) != 0;
 
 		/// <summary>
 		/// Get/Set flag indicating if entry is encrypted.
@@ -278,21 +272,8 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// <remarks>This is an assistant that interprets the <see cref="Flags">flags</see> property.</remarks>
 		public bool IsCrypted
 		{
-			get
-			{
-				return (flags & 1) != 0;
-			}
-			set
-			{
-				if (value)
-				{
-					flags |= 1;
-				}
-				else
-				{
-					flags &= ~1;
-				}
-			}
+			get => this.HasFlag(GeneralBitFlags.Encrypted);
+			set => this.SetFlag(GeneralBitFlags.Encrypted, value);
 		}
 
 		/// <summary>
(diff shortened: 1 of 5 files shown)


Numpsy:
Should ```ZipInputStream.CanDecompressEntry``` return false for these entries?

piksel:
> Should `ZipInputStream.CanDecompressEntry` return false for these entries?

Good point! It should. That is the most elegant way to handle it for the consumer.

src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
```diff
@@ -861,42 +722,29 @@ public long Crc
 		}
 
 		/// <summary>
-		/// Gets/Sets the compression method. Only Deflated and Stored are supported.
+		/// Gets/Sets the compression method. Only <see cref="CompressionMethod.Deflated">Deflated</see>
```
Numpsy, line 725: Not related to this actual issue, but that's out of date - BZip2 is supported in the ZipFile case now

src/ICSharpCode.SharpZipLib/Zip/ZipEntryExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ICSharpCode.SharpZipLib.Zip
+{
+	/// <summary>
+	/// General ZipEntry helper extensions
+	/// </summary>
+	public static class ZipEntryExtensions
```
Numpsy, line 10: Does it need to be public? (not sure if it's useful for other users or not?)

piksel: Maybe not, but I know that I would appreciate having it when consuming the library (since the flag is exposed in a public property as an `int`).

src/ICSharpCode.SharpZipLib/Zip/ZipEntry.cs
piksel: Yeah, that's the problem with having this logic in `ZipEntry`. I wonder if we even should limit this here. It should be up to `ZipFile` and `ZipOutputStream` to throw on adding such an entry.

Numpsy: refs #500

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -512,11 +512,13 @@ private int ReadingNotSupported(byte[] destination, int offset, int count)
 		/// <returns>The actual number of bytes read.</returns>
 		private int InitialRead(byte[] destination, int offset, int count)
 		{
-			if (!CanDecompressEntry)
+			if (entry == null)
```
Numpsy, line 515: I don't think it should be possible to get here in that case? (maybe  auseful sanity check in case though)

piksel: No, I don't think so either, but since the `CanDecompressEntry` used to check it, I didn't want to remove it since the rest of the body dereferences entity a lot. It shouldn't be too hard to verify though.

Numpsy: I was going to query if changing this effects the handling of AES entries (where the 'entry.IsCrypted' path only handles ZipCrypto entries), but I don't *think* it can get this far for those because of the earlier checks in GetNextEntry.

piksel: Yeah, exactly. There is only one reference to the method, and that method sets and dereferences entity just before invoking this one. Hence I removed the check.

Numpsy:
On another trivial not-directly-related-to-this-issue thought:

looking at the code just made me wonder if there enough instances of

```
if (entry.AESKeySize > 0)
```

use purely as a 'should the entry be AES encrypted' check to make it worth adding something like a

```
internal bool IsAESEncrypted => _aesEncryptionStrength > 0;
```

property to ZipEntry, even if only as an internal helper?

Numpsy:
Changes seem ok otherwise though

piksel:
Note that the reason for it immediately throwing in `InitialRead` (by invoking `StoredDescriptorEntry`) AND setting the read method to `StoredDescriptorEntry` is that in this situation, it's not just reading the entry data that is not possible. It's also not possible to skip it/seek to the next entry. This does however give the consumer a chance to read everything up to this entry and detect that such an entry was found. Perhaps not something anyone will use, but it feels like the "correct" way to handle it at least.

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -503,10 +512,7 @@ private int ReadingNotSupported(byte[] destination, int offset, int count)
 		/// <returns>The actual number of bytes read.</returns>
 		private int InitialRead(byte[] destination, int offset, int count)
 		{
-			if (!CanDecompressEntry)
-			{
-				throw new ZipException("Library cannot extract this entry. Version required is (" + entry.Version + ")");
-			}
+			var usesDescriptor = (entry.Flags & (int)GeneralBitFlags.Descriptor) != 0;
```
Numpsy, line 515: Can that use the HasFlag helper you added?

piksel: There's a lot of places where it can be used, but this is essential the same as using that method. It might be a bit easier to read, but the main issue I had were the places where the flags where compared to a magic number instead. Really opaque to the reader.
Anyhow, I won't add any more changes of that type to this PR, but I'll instead merge it so that it can be used in subsequent PRs.

PR #470, opened 2020-05-30
Experiment with using ITaggedData for AES extra data handling

Numpsy:
This is an experiment to use the ITaggedData machinery for handling the AES extra data, too see how it might effect things, and to think if it might be useful for #443 (to have the extra data handling in a central place).

Notes:

1) It changes the TagID property on ITaggedData from ```short``` to ```ushort``` to accomodate the 0x9901 value for the AES extra data. Some of the other apis use int rather than short so that might be more consistent, not sure if other places could be changed to keep it working with ```short``` though.

2) Not sure if using ZipHelperStream to unpack the data instead of reading array values is more extra overhead than would be ideal, though i think the read functions in ```ZipExtraData``` might already do more size checks than they really need to, so perhaps it could be tuned in general?

3) Similar on the ```GetData()``` side - building the fixed 7 byte array with a temporary stream and 7 calls to write byte is perhaps not the optimal approach there.

_I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the SharpZipLib open source product (the "Contribution"). My Contribution is licensed under the MIT License._

PR #457, opened 2020-04-26
add some basic async unit tests for the inflator/deflator streams

Numpsy:
refs #223 

Basically just adds async versions of the 'InflateDeflateZlib' and 'InflateDeflateNonZlib' tests and shuffles some things about to share some of the test code instead of duplicating it all.

Also changes the tests to use the NUnit 'Range' attribute for testing the different compression levels instead of having one test that loops over all the levels internally (makes the test a bit simpler, also makes each level appear as a seperate test in the results)

_I certify that I own, and have sufficient rights to contribute, all source code and related material intended to be compiled or integrated with the source code for the SharpZipLib open source product (the "Contribution"). My Contribution is licensed under the MIT License._

PR #547, opened 2020-12-07
Add async support on ZipOutputStream

### LINKED ISSUE #223: feature request: full async/await pattern stream support
### Expected behavior
I could not find any related topics, except an unanswered issue "CopyToAsync fails with ZipOutputStream #167".

One of c#'s greatest features over other languages is the async/await pattern. The .net streams all support async/await since ver 4.5 and there could be major implications and performance enhancers for FileStream in particular. Is there any plan to implement this support in this library?

0xced:
This is a very rough first pass at adding async support to ZipOutputStream.

All method doing synchronous stream writes have been duplicated with an asynchronous counterpart, starting from `ZipOutputStream.PutNextEntryAsync()`. Unfortunately, this makes a lot of duplicated code but is unavoidable to have both sync + async code paths.

.NET Standard 2.1 has been added to the target frameworks. It's required to get full async support with `Stream.DisposeAsync()`.

New unit tests must be written to cover the new async code paths.

Fixes #223

### DIFF
--- a/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
+++ b/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net45</TargetFrameworks>
     <SignAssembly>True</SignAssembly>
     <AssemblyOriginatorKeyFile>../../assets/ICSharpCode.SharpZipLib.snk</AssemblyOriginatorKeyFile>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
@@ -40,5 +40,5 @@ Please see https://github.com/icsharpcode/SharpZipLib/wiki/Release-1.3.1 for mor
       <PackagePath>images</PackagePath>
     </None>
 	 </ItemGroup>
-
+	 
 </Project>
--- a/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/DeflaterOutputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/DeflaterOutputStream.cs
@@ -2,6 +2,8 @@ using ICSharpCode.SharpZipLib.Encryption;
 using System;
 using System.IO;
 using System.Security.Cryptography;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 {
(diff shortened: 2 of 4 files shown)


Numpsy:
>>> New unit tests must be written to cover the new async code paths.

I started adapting some of the existing sync unit tests in #457 (only the deflate ones to start with) with the idea that those could be a base for testing async changes later (as there can be tests that call the async stream functions before implementing said async functions in the zip/gzip streams), but there is more work to do there for the zip/gzip subclass streams.


>>> this makes a lot of duplicated code  

hopefully there is some scope for refactoring things to avoid duplicating everything!

Numpsy:
.NET Standard 2.1 is perhaps something that can be considered seperately, as you can add it seperately from the other code changes, and that also opens up the possibility of doing things with the Span based apis and such as well?

0xced:
> .NET Standard 2.1 is perhaps something that can be considered separately

I don't think it's possible to consider it separately from the async stream feature because [Stream.DisposeAsync()](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.disposeasync) was introduced in .NET Standard 2.1.

If you want to totally avoid synchronous stream writes (typically for writing to an ASP.NET Core HTTP response body), you **must** also support `DisposeAsync()` (through `await using` in this example) which must be used instead of `Dispose()` to finish writing the zip stream asynchronously:

```csharp
await using var zipStream = new ZipOutputStream(httpContext.Response.Body);
await zipStream.PutNextEntryAsync(…);
```

See also [Add async dispose support to ZipArchive](https://github.com/dotnet/runtime/issues/1560) which is the exact same issue but for the built-in `ZipArchive` class.

0xced:
> hopefully there is some scope for refactoring things to avoid duplicating everything!

There's some room for improvement there, especially in methods which do several `WriteLE…` in a row where this could be extracted in a new method writing to a stream and then that stream would be `CopyTo` or `CopyToAsync` to the destination.

I just wanted to have a quick working proof of concept so duplicating everything and adding async support was faster without refactoring everything. I used a modified version of https://github.com/StephenClearyExamples/AsyncDynamicZip to ensure that the whole chain was actually async. I'll try to write some tests with an *AsyncOnlyStream* to ensure that no synchronous call was left.

Numpsy:
> I don't think it's possible to consider it separately from the async stream feature because [Stream.DisposeAsync()](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.disposeasync) was introduced in .NET Standard 2.1.

I mean that you can add a 2.1 target to the build/package without making any other changes, so a seperate PR could be done first to keep that change on its own.

>  several WriteLE… in a row

Think I've wondered in the past (in #470 perhaps) if doing many sequential writes for a small amount of fixed size datais the best approach, or if it could be assembled into memory first and then written into the stream in one go

0xced:
> I mean that you can add a 2.1 target to the build/package without making any other changes, so a separate PR could be done first to keep that change on its own.

Oh I see, I misunderstood what you meant. I can do a pull request with just that: adding the **.NET Standard 2.1** target framework.

> Think I've wondered in the past (in #470 perhaps) if doing many sequential writes for a small amount of fixed size datais the best approach, or if it could be assembled into memory first and then written into the stream in one go

This should probably be benchmarked to make sure what approach is the most performant. Let's hope that writing in memory first is more performant so that we can win on both maintainability and performance!

piksel:
Yeah, basically everything that is not actual compressed data should probably just be written to a buffer and then piped asynchronously or not to the base stream.
This would allow us to keep the duplicated code minimal and probably not lose any noticeable performance.
Calls to `WriteLEShortAsync` is probably not ideal to do anyway, since the stack overhead of that call is larger than the actual data needing to be buffered.

Numpsy:
This might be getting too far ahead of things, but something to consider in case - The way that ZipOutputStream and GZipOutputStream both derive from DeflatorOutputStream could possibly cause issues if Async functions are added in DeflatorOutputStream but not GZipOutputStream (e.g. does adding a FinishAsync to DeflatorOutputStream mean that someone could call that via an instance of GZipOutputStream and have it do the wrong thing?)

piksel:
@Numpsy If we don't implement them, I guess they should be made virtual and overridden by `throw NotImplementedException()` and a comment about them not being available for GZip yet.

0xced:
Another possible approach to avoid code duplication is using a t4 text template to generate both sync and async versions of the same code. I saw this today in the [CsvHelper library](https://github.com/JoshClose/CsvHelper/blob/ff76cd2895bd740ee872de1ffbae6367ad87441c/src/CsvHelper/CsvParser.tt#L83-L108).

Not sure what is the most maintainable...

piksel:
I think this a pretty good use case for meta programming. Too bad the tooling support for T4 is terrible. Maybe that has changed though, haven't tried it with OmniSharp or Rider...

0xced:
Rider got [support for T4 Text Templates][1] in version 2019.3 and it's pretty decent.

[1]: https://blog.jetbrains.com/dotnet/2019/12/17/introducing-t4-text-template-support/

piksel:
Actually, that link pointed to another problem; github doesn't highlight them :/ 

I made an attempt at splitting the logic from the "actual" writes so that just the writes could be put in separate Sync/Async methods in #574.

The methods like `PutNextEntry` are ~250 LoC, which is really hard to maintain in two copies.

0xced:
Awesome work, Nils! I'm closing this pull request in favour of yours (much improved) in #574.

PR #549, opened 2020-12-07
Add .NET Standard 2.1 target framework

0xced:
Also fix the .NET Standard 2.0 target framework moniker: netstandard2.0 instead of netstandard2, see https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks


piksel:
Ah, that's why the CI failed. Nice!

piksel (approved):

PR #551, opened 2020-12-11
AES decryption support in ZipInputStream

remogloor:
This pull request continues the work of @Numpsy and replaces Pull Request #381 

The issue with the HMAC calculation in the original pull request was that it always decrypted and calculated HMAC for the entire input buffer. But actually it should only decrypt and calculate the HMAC up to the compressed size of the current entry. This issue is fixed now and all the unit tests pass. AES encrypted Zips can now be decrypted and decompressed using the ZipInput Stream.

Original Information from #381

> I started looking at this and then haven't had time to try to finish it yet, so here it is in case anyone has any ideas.
> 
> The basic decryption seems to be ok, but the auth code checking is disabled as it doesn't work. I think the issue is the way that inflater input buffer runs all its data through the crypto transform rather than just the part which is actually encrypted, so the auth code calculation in the AES transform gets the wrong result.
> Not sure of the best way to handle it - might need changes in the input buffer, or a different approach to calculating the value?
> 
> The changes also have a couple of more minor changes in separate commits, in case they're useful on their own.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs
+++ b/src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs
@@ -8,8 +8,6 @@ namespace ICSharpCode.SharpZipLib.Encryption
 	/// </summary>
 	internal class ZipAESTransform : ICryptoTransform
 	{
-		private const int PWD_VER_LENGTH = 2;
-
 		// WinZip use iteration count of 1000 for PBKDF2 key generation
 		private const int KEY_ROUNDS = 1000;
 
@@ -28,6 +26,7 @@ namespace ICSharpCode.SharpZipLib.Encryption
 		private byte[] _authCode = null;
 
 		private bool _writeMode;
+		private Action<int> _appendHmac = remaining => { };
 
 		/// <summary>
 		/// Constructor.
@@ -63,11 +62,28 @@ namespace ICSharpCode.SharpZipLib.Encryption
 
 			// Use empty IV for AES
 			_encryptor = rm.CreateEncryptor(key1bytes, new byte[16]);
-			_pwdVerifier = pdb.GetBytes(PWD_VER_LENGTH);
+			_pwdVerifier = pdb.GetBytes(Zip.ZipConstants.AESPasswordVerifyLength);
 			//
 			_hmacsha1 = IncrementalHash.CreateHMAC(HashAlgorithmName.SHA1, key2bytes);
 			_writeMode = writeMode;
 		}
+
+		/// <summary>
+		/// Append all of the last transformed input data to the HMAC.
+		/// </summary>
+		public void AppendAllPending()
+		{
+			_appendHmac(0);
+		}
+		
+		/// <summary>
+		/// Append all except the number of bytes specified by remaining of the last transformed input data to the HMAC.
+		/// </summary>
+		/// <param name="remaining">The number of bytes not to be added to the HMAC. The excluded bytes are form the end.</param>
+		public void AppendFinal(int remaining)
+		{
+			_appendHmac(remaining);
+		}		
 
 		/// <summary>
 		/// Implement the ICryptoTransform method.
@@ -78,8 +94,16 @@ namespace ICSharpCode.SharpZipLib.Encryption
 			// This does not change the inputBuffer. Do this before decryption for read mode.
 			if (!_writeMode)
 			{
-				_hmacsha1.AppendData(inputBuffer, inputOffset, inputCount);
+				if (!ManualHmac)
+				{
+					_hmacsha1.AppendData(inputBuffer, inputOffset, inputCount);
+				}
+				else
+				{
+					_appendHmac = remaining => _hmacsha1.AppendData(inputBuffer, inputOffset, inputCount - remaining);
+				}
 			}
+
 			// Encrypt with AES in CTR mode. Regards to Dr Brian Gladman for this.
 			int ix = 0;
 			while (ix < inputCount)
@@ -167,6 +191,13 @@ namespace ICSharpCode.SharpZipLib.Encryption
 		/// Gets a value indicating whether the current transform can be reused.
 		/// </summary>
 		public bool CanReuseTransform => true;
+
+		/// <summary>
+		/// Gets of sets a value indicating if the HMAC is updates on every read of if updating the HMAC has to be controlled manually
+		/// Manual control of HMAC is needed in case not all the Transformed data should be automatically added to the HMAC.
+		/// E.g. because its not know how much data belongs to the current entry before the data is decrypted and analyzed. 
+		/// </summary>
+		public bool ManualHmac { get; set; }
 
 		/// <summary>
 		/// Cleanup internal state.
--- a/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Security.Cryptography;
+using ICSharpCode.SharpZipLib.Encryption;
 
 namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 {
@@ -92,8 +93,24 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 		public int Available
 		{
 			get { return available; }
-			set { available = value; }
+			set
+			{
+				if (cryptoTransform is ZipAESTransform ct)
+				{
+					ct.AppendFinal(value);
+				}
+				
+				available = value;
+			}
 		}
+
+		/// <summary>
+		/// A limitation how much data is decrypted. If null all the data in the input buffer will be decrypted.
+		/// Setting limit is important in case the HMAC has to be calculated for each zip entry. In that case
+		/// it is not possible to decrypt all available data in the input buffer, and only the data
+		/// belonging to the current zip entry must be decrypted so that the HMAC is correctly calculated. 
+		/// </summary>
+		internal int? DecryptionLimit { get; set; }
 
 		/// <summary>
 		/// Call <see cref="Inflater.SetInput(byte[], int, int)"/> passing the current clear text buffer contents.
@@ -113,6 +130,11 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 		/// </summary>
 		public void Fill()
 		{
+			if (cryptoTransform is ZipAESTransform ct)
+			{
+				ct.AppendAllPending();
+			}
+
 			rawLength = 0;
 			int toRead = rawData.Length;
 
@@ -127,13 +149,11 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 				toRead -= count;
 			}
 
+			clearTextLength = rawLength;
 			if (cryptoTransform != null)
 			{
-				clearTextLength = cryptoTransform.TransformBlock(rawData, 0, rawLength, clearText, 0);
-			}
-			else
-			{
-				clearTextLength = rawLength;
+				var size = CalculateDecryptionSize(rawLength);
+				cryptoTransform.TransformBlock(rawData, 0, size, clearText, 0);
 			}
 
 			available = clearTextLength;
@@ -290,7 +310,9 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 					clearTextLength = rawLength;
 					if (available > 0)
 					{
-						cryptoTransform.TransformBlock(rawData, rawLength - available, available, clearText, rawLength - available);
+						var size = CalculateDecryptionSize(available);
+
+						cryptoTransform.TransformBlock(rawData, rawLength - available, size, clearText, rawLength - available);
 					}
 				}
 				else
@@ -299,6 +321,19 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 					clearTextLength = rawLength;
 				}
 			}
+		}
+
+		private int CalculateDecryptionSize(int availableBufferSize)
+		{
+			int size = DecryptionLimit ?? availableBufferSize;
+			size = Math.Min(size, availableBufferSize);
+			
+			if (DecryptionLimit.HasValue)
+			{
+				DecryptionLimit -= size;
+			}
+
+			return size;
 		}
 
 		#region Instance Fields
@@ -459,9 +494,10 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 		/// <summary>
 		/// Clear any cryptographic state.
 		/// </summary>
-		protected void StopDecrypting()
+		protected virtual void StopDecrypting()
 		{
 			inputBuffer.CryptoTransform = null;
+			inputBuffer.DecryptionLimit = null;
 		}
 
 		/// <summary>
--- a/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
@@ -325,13 +325,13 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 
 		private int CalculateDecryptionSize(int availableBufferSize)
 		{
-			int size = DecryptionLimit ?? availableBufferSize;
-			size = Math.Min(size, availableBufferSize);
-			
-			if (DecryptionLimit.HasValue)
+			if (!DecryptionLimit.HasValue)
 			{
-				DecryptionLimit -= size;
+				return availableBufferSize;
 			}
+			
+			var size = Math.Min(DecryptionLimit.Value, availableBufferSize);
+			DecryptionLimit -= size;
 
 			return size;
 		}
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
@@ -74,10 +74,10 @@ namespace ICSharpCode.SharpZipLib.Zip
 		private ZipEntry entry;
 
 		private long size;
-		private CompressionMethod method;
 		private int flags;
 		private string password;
 		private readonly StringCodec _stringCodec = ZipStrings.GetStringCodec();
+		private ZipAESTransform cryptoTransform;
 
 		#endregion Instance Fields
 
@@ -141,17 +141,22 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// <summary>
 		/// Is the compression method for the specified entry supported?
 		/// </summary>
-		/// <remarks>
-		/// Uses entry.CompressionMethodForHeader so that entries of type WinZipAES will be rejected. 
-		/// </remarks>
 		/// <param name="entry">the entry to check.</param>
 		/// <returns>true if the compression method is supported, false if not.</returns>
 		private static bool IsEntryCompressionMethodSupported(ZipEntry entry)
 		{
-			var entryCompressionMethod = entry.CompressionMethodForHeader;
+			var entryCompressionMethodForHeader = entry.CompressionMethodForHeader;
+			var entryCompressionMethod = entry.CompressionMethod;
 
-			return entryCompressionMethod == CompressionMethod.Deflated ||
-				   entryCompressionMethod == CompressionMethod.Stored;
+			var compressionMethodSupported = 
+				entryCompressionMethod == CompressionMethod.Deflated ||
+				entryCompressionMethod == CompressionMethod.Stored;
+			var entryCompressionMethodForHeaderSupported = 
+				entryCompressionMethodForHeader == CompressionMethod.Deflated ||
+				entryCompressionMethodForHeader == CompressionMethod.Stored ||
+				entryCompressionMethodForHeader == CompressionMethod.WinZipAES;
+
+			return compressionMethodSupported && entryCompressionMethodForHeaderSupported;
 		}
 
 		/// <summary>
@@ -191,7 +196,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 			var versionRequiredToExtract = (short)inputBuffer.ReadLeShort();
 
 			flags = inputBuffer.ReadLeShort();
-			method = (CompressionMethod)inputBuffer.ReadLeShort();
+			var method = (CompressionMethod)inputBuffer.ReadLeShort();
 			var dostime = (uint)inputBuffer.ReadLeInt();
 			int crc2 = inputBuffer.ReadLeInt();
 			csize = inputBuffer.ReadLeInt();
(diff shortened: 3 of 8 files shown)


Numpsy:
One of the issues here is the way that the inflator input machinery works by trying to consume all the input so that it can work in cases where it doesn't know the length of the data in advance (i.e. when reading a Zip entry with a descriptor).
In that case, it doesn't know what value ```DecryptionLimit ``` should be, so it can't set it.

I've thought in the past that it could be changed to use the compressed size fields in cases where they're known, but that could be a more generic change and not encryption specific.

Saying that - just being able to decrypt entries that don't use a descriptor is still a benefit, just needs a few extra checks to make sure things like CanDecompress give the correct values.

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -148,7 +148,8 @@ private static bool IsEntryCompressionMethodSupported(ZipEntry entry)
 			var entryCompressionMethod = entry.CompressionMethodForHeader;
 
 			return entryCompressionMethod == CompressionMethod.Deflated ||
-				   entryCompressionMethod == CompressionMethod.Stored;
+				   entryCompressionMethod == CompressionMethod.Stored ||
+				   entryCompressionMethod == CompressionMethod.WinZipAES;
```
Numpsy, line 152: I don't think this is right - in the AES case it needs to check what the real compression method is, otherwise something like encrypted BZip2 would be included. (I think I only made it use CompressionMethodForHeader to exclude AES, if it does support AES then CompressionMethod should do?)

remogloor:
> One of the issues here is the way that the inflator input machinery works by trying to consume all the input so that it can work in cases where it doesn't know the length of the data in advance (i.e. when reading a Zip entry with a descriptor).
> In that case, it doesn't know what value `DecryptionLimit ` should be, so it can't set it.
> 
> I've thought in the past that it could be changed to use the compressed size fields in cases where they're known, but that would be a more generic change and not encryption specific.
> 
> Saying that - just being able to decrypt entries that don't use a descriptor is still a benefit, just needs a few extra checks to make sure things like CanDecompress give the correct values.
> 
> (/not particularly a fan of more zip specific encryption code in the inflator, but that's how it works so, !)

I don't quite get what situation you are refering to.

To my understanding the Zip specification ensures that a Zip entry always comes with a header that specifies the compressed size. The `DecryptionLimit` is actually set to that compressed size as this is exactly what we want to be used here. The compressed data is decrypted and HMAC is calculated on the compressed data.  After that happend the inflator can inflate the data. The uncompressed size does not need to be known in advance. (even though that is part of the entry header too.)

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
remogloor: Changed this to check that the entry.CompressionMethod is also Deflated or Stored.

Numpsy:
> 
> To my understanding the Zip specification ensures that a Zip entry always comes with a header that specifies the compressed size. 

Not quite - there are two modes of operation for entries - one which puts the sizes in the file header, and one which sets the header values to 0 and puts the sizes in a ```Descriptor``` field which is after the file data. (it's described in the [AppNote](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT), e.g in sections 4.4.8 and 4.3.9).
In those cases ZipInputStream has to read the data before it can know the length (that;s also the reason why it can't read entries which have a descriptor and are stored rather than compressed).

remogloor:
> Not quite - there are two modes of operation for entries - one which puts the sizes in the file header, and one which sets the header values to 0 and puts the sizes in a `Descriptor` field which is after the file data. (it's described in the [AppNote](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT), e.g in sections 4.4.8 and 4.3.9).

Ok got it. That use case is now covered and tested as well.
Although it seems that this is really an edge case as non of the big players actually supports this use case:

- Linux zip => no support to create AES encrypted files
- Pkzip => no support to create AES encrypted files
- 7Zip => no support to create files with data descruptor
- WinZip => no support to create files with data descruptor

The only way I could create such a file that has both AES encryption and data descriptor was using SharpZipLib itself.

piksel:
Well, the reason is that Descriptors are used for streaming data, where you might not know the file size before hand, or you cannot modify the stream to update the header (like with a web server response). This scenario doesn't really make sense with a file-based stand-alone executable. Perhaps for a CLI version that pipes to stdout, not sure if 7z CLI might do it perhaps?

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
```diff
@@ -544,6 +550,17 @@ public void CreateZipWithEncryptedEntries(string password, int keySize, Compress
 			}
 		}
 
+		private class NonSeekableMemoryStream : MemoryStream
```
Numpsy, line 553: trivial, but: there is already a ```MemoryStreamWithoutSeek``` class that does that in the tests?

piksel: yeah, ` ICSharpCode.SharpZipLib.Tests.TestSupport.MemoryStreamWithoutSeek`

remogloor:
Ready for review

Numpsy:
I'll try to have a look later

Numpsy:
Did I ever query whether there would be any benefir of making use of the entry size (when known) in all cases, rather than just for AES entries (e.g. something like what DecryptionLimit does, but for plain compressed entries)?

src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
```diff
@@ -92,9 +93,25 @@ public byte[] ClearText
...
+			}
 		}
 
+		/// <summary>
+		/// A limitation how much data is decrypted. If null all the data in the input buffer will be decrypted.
+		/// Setting limit is important in case the HMAC has to be calculated for each zip entry. In this case
+		/// it is not possible to decrypt all available data in the input buffer. In this case only the data
+		/// belonging to the current zip entry must be decrypted so that the HMAC is correctly calculated. 
+		/// </summary>
+		internal int? DecryptionLimit { get; set; }
```
Numpsy, line 113: Does this need to be set back to null between entries? (you might have a zip file with an AES encrypted entry, followed by a ZipCrypto or unencrypted entry, and you wouldn't want the limit set for the second entry?)

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -522,27 +576,67 @@ private int InitialRead(byte[] destination, int offset, int count)
...
+					byte[] pwdVerifyCalc = decryptor.PwdVerifier;
+					if (pwdVerifyCalc[0] != pwdVerifyRead[0] || pwdVerifyCalc[1] != pwdVerifyRead[1])
+					{
+						throw new ZipException("Invalid password for AES");
+					}
+
+					// The AES data has saltLen+AESPasswordVerifyLength bytes as a header, and AESAuthCodeLength bytes
+					// as a footer.
+					csize -= (saltLen + ZipConstants.AESPasswordVerifyLength + ZipConstants.AESAuthCodeLength);
+					inputBuffer.DecryptionLimit = (int)csize;
```
Numpsy, line 637: Do we want to set DecryptionLimit in cases where we don't know the compressed size? (looks like CalculateDecryptionSize might keep subtracting values from it, even if it's negative to begin with?)

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -697,6 +791,8 @@ private int BodyRead(byte[] buffer, int offset, int count)
 						}
 					}
 					break;
+				default:
+					throw new InvalidOperationException("Internal Error: Unsupported compression method encountered.");
```
Numpsy, line 795: do we ever allow it to get to this point for an entry with an unsupported compression method? (maybe it makes sense to check in any case)

Numpsy:
(I still need to have a more detailed read over the other logic in the nexzt day or two)

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
remogloor: No it should never get to this point.

This is just precaution in case there would ever an additional compression method so that it's not forgotten to be extended here.

src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
remogloor: It's not really necessary as the CryptoTransform is set to null and DecryptionLimit is only considered when CryptoTransform is not null.

But still I changed this to be reset to null.

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
remogloor: If it is set to a negative value it is behaving the same as when it is set to null.

But I set it to null now and removed the check for a negative value in the InflaterInputStream instead.

remogloor:
> I'll try to have a look later
> 
> > This scenario doesn't really make sense with a file-based stand-alone executable
> 
> I've seen document files created by LibreOffice (using the file formats that use zip file containers) that use descriptors, though probably not together with this form of encryption (just as an example of how these sort of files can come about through 'normal' file based tools)

It doesn't matter how these entries can be created with other libraries. It is possible to create them with SharpZipLib. And there is a test to verify these entries can be unzipped and decrypted.

remogloor:
Is there any update for this pull request?
Is something needed so that it can get merged?

Numpsy:
Sorry, I haven't had much spare time to do things lately, I'll have another look over the weekend

Numpsy:
Is that last change a separate bug fix for ZipOutputStream?

remogloor:
Yes and No

It's a separate bugfix for the ZipOutputStream.

But due to that bug it is not possible to create certain zip files needed fo the unit tests of this pull request.
The tests fail without this change as the zip files created during setup of the unit tests are not according to the specification.

piksel:
Regarding `ZipOutputStream`:

The specification says that the CRC should be set to 0, and only for AE-2 and explicitly states that
>NOTE: Zip utilities that support the AE-2 format are required to be able to read files that were created in the AE-1 format, and during decryption/extraction of files in AE-1 format should verify that the file's CRC matches the value stored in the CRC field.

And I think that 
>  the standard Zip CRC value is not used  

should be interpreted as "when reading, ignore the Zip CRC", it shouldn't refuse to read it.
And the same goes for the descriptor, even if it's not strictly needed, it shouldn't cause any exceptions if it's there. I don't really see what test you were fixing with this either, as the commit before it had all tests passing (except for `InvalidPasswordSeekable` which may randomly fail due to the random zip generated may lead to invalid dist codes being found with the wrong password)?

remogloor:
Seems I had the issue not correctly in mind anymore. The problem is not that the tests fail. But that the tests do not test what they are supposed to test.

There are multiple use cases that need to be tested. AES decryption needs to be able to handle both cases the one where there is a data descrpytor and the one where there isn't one. This is done by one tests that used InlineData to test the different variations.

`public void ZipInputStreamDecryption(int aesKeySize, CompressionMethod compressionMethod, bool forceDataDescriptor, bool skipEntries, bool partiallyReadEntries)`

The thing is here that the parameter forceDataDescriptor is basically meaningless if the ZipOutputStream always adds one anyways in case there is a password. With the unchanged ZipOutputStream it is only possible to test the cases where there is a data descryptor as only such files can be created by it. All the cases where there isn't a data desscryptor can't be tested otherwise. 

With my second last change I fixed a bug that lead to an exception in case an entry was only read partially and not until the end of the stream. If you undo these changes (only the ZipInputStream not the tests) and the changes to the ZipOutputStream you will notice that only the test case that "stores" entries will fail. The test case that "deflate" the entries will pass successfully which is not what is expected. But if you change the ZipOutputStream so that there is no data descryptor when it is not needed it will fail as expected. 

By the way, tools like 7-Zip are also not adding a data descryptor in this case.

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
```diff
@@ -126,40 +127,80 @@ public void ZipFileAesDecryption()
...
+		[TestCase(128, CompressionMethod.Stored,   false, false, true)]
+		[TestCase(256, CompressionMethod.Deflated, false, false, true)]
+		[TestCase(256, CompressionMethod.Stored,   false, false, true)]
+		[TestCase(0,   CompressionMethod.Deflated, true,  false, true)]
+		[TestCase(0,   CompressionMethod.Stored,   true,  false, true)]
+		[TestCase(128, CompressionMethod.Deflated, true,  false, true)]
+		[TestCase(128, CompressionMethod.Stored,   true,  false, true)]
+		[TestCase(256, CompressionMethod.Deflated, true,  false, true)]
+		[TestCase(256, CompressionMethod.Stored,   true,  false, true)]		
+		public void ZipInputStreamDecryption(int aesKeySize, CompressionMethod compressionMethod, bool forceDataDescriptor, bool skipEntries, bool partiallyReadEntries)
```
piksel, line 168: This is getting out hand...
Perhaps use [ValuesAttribute](https://docs.nunit.org/articles/nunit/writing-tests/attributes/values.html) instead?

piksel (commented):
Yeah, not adding it when not needed is indeed the right thing to do, I just questioned not allowing them to be there. 

I think I understand what happened, but my internal zip emulation is at its limits 😁.
Just a note on the test arguments...

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
remogloor: I can change that although it would add the combination skipEntries = true & partiallyReadEntries = true which is not really needed and would therefore result in 12 additional unnecessary test cases.

piksel:
```cs
enum EntryMode { Skip, PartiallyRead, FullRead }
```
It's not readable the way it is right now. And how would you know if you missed a combination?

piksel: Actually, I think these should be separate tests all together. It seems like you are testing different scenarios.
Just copy the whole test and make changes to it with an updated description and name, the encryption level is probably not necessary to have all variants of, but include it if you disagree.

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
```diff
@@ -108,6 +120,101 @@ public void ZipFileAesDecryption()
...
+						if (skipEntries && fileNumber % 2 == 1)
+						{
+							continue;
+						}
+						
+						if (partiallyReadEntries && fileNumber % 2 == 1)
+						{
+							zis.ReadByte();
+							continue;
+						}
```
piksel, line 204: I think you are missing the point a little. This part is not good for a test. You have several tests in the same body and and switching them out.
And I still have no idea why you are checking "short" entries for some tests and not for others.
If some tests have a common setup/teardown, that part *can* be moved to a helper function, but it's also fine to just have several copies of the same code.
The test should clearly test for a single thing, not require you to resolve the values of bools to check. This means that the reviewers and subsequent developers can look at the test and directly see what the scope and purpose of the test is.

Also, as a general rule, don't have bools at all in helpers, and if you do, always call them with the name (like the StreamReader just below). That way you don't have to consult the docs or look up the function to understand what the purpose is.

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
```diff
@@ -434,40 +541,77 @@ public void ZipFileAESReadWithEmptyPassword()
...
 				}
 			}
 		}
+		
+		/// <summary>
+		/// Test reading an AES encrypted file and reading some entries only partially be not reading to the end of stream.
+		/// </summary>
+		[Test]
+		[Category("Zip")]
+		public void ZipFileAESReadEntriesCanBeReadPartially()
```
piksel, line 631:
```suggestion
		public void ZipInputStreamAESReadEntriesCanBeReadPartially()
```
Same as above, `ZipFile` is another class in the library.
I guess the naming is because it's testing on a static precomputed file?

src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
```diff
@@ -92,9 +93,25 @@ public byte[] ClearText
...
+				
+				available = value;
+			}
 		}
 
+		/// <summary>
+		/// A limitation how much data is decrypted. If null all the data in the input buffer will be decrypted.
+		/// Setting limit is important in case the HMAC has to be calculated for each zip entry. In this case
+		/// it is not possible to decrypt all available data in the input buffer. In this case only the data
+		/// belonging to the current zip entry must be decrypted so that the HMAC is correctly calculated.
```
piksel, line 111:
```suggestion
		/// Setting limit is important in case the HMAC has to be calculated for each zip entry. In that case
		/// it is not possible to decrypt all available data in the input buffer, and only the data
		/// belonging to the current zip entry must be decrypted so that the HMAC is correctly calculated. 
```

piksel (changes requested):

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -326,13 +342,51 @@ private void ReadDataDescriptor()
...
+
+			if (testAESAuthCode && entry.AESKeySize != 0)
+			{
+				byte[] authBytes = new byte[ZipConstants.AESAuthCodeLength];
+				int authBytesRead = inputBuffer.ReadRawBuffer(authBytes, 0, authBytes.Length);
+
+				if (authBytesRead < ZipConstants.AESAuthCodeLength)
+				{
+					throw new ZipException("Internal error missed auth code"); // Coding bug
+																			// Final block done. Check Auth code.
```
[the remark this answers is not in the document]
remogloor: Fixed

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
```diff
@@ -108,6 +120,101 @@ public void ZipFileAesDecryption()
...
+						{
+							zis.ReadByte();
+							continue;
+						}
+
+						using (var sr = new StreamReader(zis, Encoding.UTF8, leaveOpen: true, detectEncodingFromByteOrderMarks: true, bufferSize: 1024))
+						{
+							var content = sr.ReadToEnd();
+
+							Assert.AreEqual(fileNumber < 3 ? DummyDataString : DummyDataStringShort, content,
```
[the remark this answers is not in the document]
remogloor: Reduced size.
The size needs to be well above the size of the InflaterInputStream buffer size as the behavior changes at certain places depending if the file fits into that buffer or not.

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
remogloor: Sepatarted into several tests

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
remogloor: Split into several tests.

I check here for different sizes of entries as I figured out that the behavior internally is different in case of AES depending if the entry fits into the buffer of the InflaterInputStream or not. If I have it correctly in mind then it is about the calculation of the AES Auth code.

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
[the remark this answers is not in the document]
piksel: ...which is what the short and long variants are for, I presume?
That explanation above should really be in the tests to show the significance of the buffer-/data- sizes, perhaps by using a constant for the buffer size and comments for the dummy strings which explain that they are larger than the buffer.

Perhaps even use something like:
```cs
// Random data that is smaller than the size of the read buffer
SmallDummyData = byte[ReadBufferSize / 2];
random.GetBytes(SmallDummyData);

// Random data that is larger than the size of the read buffer
LargeDummyData = byte[ReadBufferSize * 2];
random.GetBytes(LargeDummyData);
```

And perhaps something like this to explain what index 5 of the name should indicate.
```cs
bool entryIsSmall = entry.Name[5] > 3;
```

src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
```diff
@@ -301,6 +323,19 @@ public ICryptoTransform CryptoTransform
...
+		{
+			int size = DecryptionLimit ?? availableBufferSize;
+			size = Math.Min(size, availableBufferSize);
+			
+			if (DecryptionLimit.HasValue)
+			{
+				DecryptionLimit -= size;
+			}
+
+			return size;
```
piksel, line 336:
```suggestion
			if (!DecryptionLimit.HasValue)
			{
			    return availableBufferSize;
			}

			var size = Math.Min(DecryptionLimit, availableBufferSize);
			DecryptionLimit -= size;

			return size;
```

Does look simpler to me, as three of the operations only apply when it's not null.

src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs
```diff
@@ -88,6 +89,23 @@ public ZipAESTransform(string key, byte[] saltBytes, int blockSize, bool writeMo
...
+			_appendHmac(0);
+		}
+		
+		/// <summary>
+		/// Append all except the number of bytes specified by remaining of the last transformed input data to the HMAC.
+		/// </summary>
+		/// <param name="remaining">The number of bytes not to be added to the HMAC. The excluded bytes are form the end.</param>
+		public void AppendFinal(int remaining)
+		{
+			_appendHmac(remaining);
```
piksel, line 107: This delegate is only set if in Manual mode, otherwise it's just a nop, right? Why not just bail inside the method if not `ManualHmac` is set?
It just seems like added complexity, or am I missing something?

piksel (commented):
I think I get how it works now, more or less. These where just some thoughts when trying to follow the exciting journey of `DecryptionLimit`.

pmpbpl:
Hi, any updates?

src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs
remogloor: Yes you are missing that the delegate captures the information when
   TransformBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
       _appendHmac = remaining => _hmacsha1.AppendData(inputBuffer, inputOffset, inputCount - remaining);
is called. During AppendFinal and AppendAllPending that information is not available anymore.

src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
remogloor: Changed

test/ICSharpCode.SharpZipLib.Tests/Zip/ZipEncryptionHandling.cs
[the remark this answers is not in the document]
remogloor: Changed

lukasweber:
Hi @piksel, is there any chance you could review the changes from @remogloor anytime soon? We would be very happy to have this feature in the library.

piksel (approved):

PR #554, opened 2020-12-19
skip CRC calculation for AES zip entries

### LINKED ISSUE #552: AES encrypted Zips not matching specification
### Steps to reproduce
1. Create a AES encrypted zip using ZipOutputStream, probably also using ZipFile

### Expected behavior
A Zip according to the specification is created
https://www.winzip.com/win/en/aes_info.html#comp-method

The specication states that:
- AE-1: The version supports CRC, The CRC must be set to the actual CRC of the content
- AE-2: This version does not support CRC, CRC value must be 0

The reason behind not using CRC in case of AE-2 is that this is an unnecessary redundant overhead. AES has an Auth code which is even better than a simple CRC. Therefore the computation time for the CRC can be removed to improve perfromance.

### Actual behavior
A AE-2 zip is created that has the CRC value set to the actual value

To match the specification:
->  Either change the version to AE-1
-> Or set the CRC to 0

In order to fix this issue it first needs to be decided which version the zip files shall use:
AE-1 or AE-2

- Compiled from source, commit: b7bc4e08620f4410b721a3bd59fe53537f774909

piksel:
This skips AES calculation for AES encrypted entries to align with AE-2.

Fixes #552 

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -960,6 +960,9 @@ namespace ICSharpCode.SharpZipLib.Zip
 
 					if (testing && testData && this[entryIndex].IsFile)
 					{
+						// Don't check CRC for AES encrypted archives
+						var checkCRC = this[entryIndex].AESKeySize == 0;
+
 						if (resultHandler != null)
 						{
 							status.SetOperation(TestOperation.EntryData);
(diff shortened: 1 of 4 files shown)


src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -960,6 +960,9 @@ public bool TestArchive(bool testData, TestStrategy strategy, ZipTestResultHandl
 
 					if (testing && testData && this[entryIndex].IsFile)
 					{
+						// Don't check CRC for AES encrypted archives
+						var checkCRC = this[entryIndex].AESKeySize == 0;
```
Numpsy, line 964: should whether the CRC is checked on read depend on the whether the entry is AE-1 or AE-2, rather than just being sklpped for all AES entries? (e.g. dependant the _aesVer value in ZipEntry)

piksel: It could, but then it would fail on archives that would otherwise be perfectly OK. The reason for it being dropped in AE-2 is that it's not necessary in AE-1.
I don't think `TestArchive` should fail because the file is technically out of spec, even though it's fully readable. This method is used in our own tests as well, but it's "main" purpose is for the consumer to verify that an archive is valid before attempting extraction etc.

PR #555, opened 2020-12-27
FastZip - fix TimeSetting control when extracting Zip entries

va4es2:
When creating Zip entries with FastZip - it is possible to control timesetting for files or dirs. But when extraction it is abcent. ZipEntry's DateTime property always set to extracted entry by SetLastWriteTime() function.


piksel:
This does look great, but would you mind adding some tests?

piksel:
Closed in favor of #566 due to lack of response.

PR #574, opened 2021-02-14
feat(zip): ZipOutputStream async support

piksel:
Based on #547, but rather than duplicate the already massive zip format writing methods, it combines all the "logic" (which isn't really async anyway) and only does the actual writing using a-/sync methods.

This also moves the byte order swapping from the odd `ZipHelperStream` into a much more lightweight (and probably faster) `ZipFormat` helper class. It should also help DRYing up the shared code in `ZipFile` and `Zip*Stream`.

No actual benchmarks have been run yet, but the tests should pass.


piksel:
@cloghead It's not the easiest thing to pick up for a new project, but some code review could be nice. 

This will likely be the focus for the next minor release.

piksel:
@Numpsy That's pretty much what I expected (not the odd memory usage ofc). The overhead that got removed was just instantiating `ZipHelperStream`, which I guess should be fairly easy to optimize out by the JIT. I'll do a rebase and compare the results.

guipcarvalho:
Hello guys, is this code going to be merged in the near future? I'm facing this issue where I need to create a zip out of a asynconlystream

Numpsy:
I haven't had a chance to look at anything for some time and have lost track of what state the async work was in, need to remind myself what was going on :-(

piksel:
@guipcarvalho That is the plan, but I have been busy with other projects. This should be ready to merge, I'll try to find some time to get it done this weekend.

guipcarvalho:
Thank you @piksel!

Numpsy:
I think I might have queried before, but don't recall the answer, so just in case - is there a need to check the behavior of GzipOutputStream with regards to the new functions on DeflaterOutputStream? (e.g. is it possible to call FinishAsync or DisposeAsync through a GzipOutputStream, and if so does doing so cause any problems?)

piksel:
@Numpsy I made a test that was simply a Async version of `GzipTests.SmallBufferDecompressionAsync`, and it seems to work fine... 🤷‍♀️

```cs
using (var gzos = new GZipOutputStream(msGzip))
{
    gzos.IsStreamOwner = false;
    
    await gzos.WriteAsync(inputBuffer, 0, inputBuffer.Length);
    await gzos.FlushAsync();
    await gzos.FinishAsync(CancellationToken.None);
}
```

Issue #572, opened 2021-02-11
.net core zip+aes ReadAsync failed on one of files


ShadowTeolog:
### Steps to reproduce
1.open archive with password "TestStorage"
2.try extract file "[www/autopilotandbridge/home.png]" using 
`var entry = _zf.GetEntry(path);
var zipStream = _zf.GetInputStream(entry);
var bufer = new byte[entry.Size];
var result = await zipStream.ReadAsync(bufer, 0, bufer.Length);`

### Expected behavior
file is etracted

### Actual behavior
Got exception
`
{"TransformFinalBlock is not implemented and inputCount is greater than 0"}
"   at ICSharpCode.SharpZipLib.Encryption.ZipAESTransform.TransformFinalBlock(Byte[] inputBuffer, Int32 inputOffset, Int32 inputCount)\r\n   at System.Security.Cryptography.CryptoStream.<ReadAsyncCore>d__43.MoveNext()\r\n   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()\r\n   at System.Security.Cryptography.CryptoStream.<ReadAsyncInternal>d__38.MoveNext()\r\n   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()\r\n   at ResourceCollection.LazyZipDirectoryLoader.`
### Obtained from (only keep the relevant lines)
- Package installed using NuGet v1.3.1
[ConsoleApp2.zip](https://github.com/icsharpcode/SharpZipLib/files/5966267/ConsoleApp2.zip)

Numpsy:
[...]

PR #576, opened 2021-02-14
[WIP] Add an implementation of ReadAsync to ZipAESStream

Numpsy:
refs #572 

Add an implementation of ReadAsync to ZipAESStream, with associated StreamUtils change and a copy/pasted unit test (that currently fails on .NETCore 3.1 due to #572).

WIP because:

1) There is still a bunch of duplication between ReadAndTransform/ReadAndTransformAsync, and the copy/pasted signature of TransformFromSlideBuffer doesn't feel ideal.

2) Looking at ZipFile.GetInputStream a bit more, I don't think it will _actually_ do async reads from the base stream unless ```PartialInputStream``` is given some async support as well, so the current code might not actually be any 'more' async that just making ```ReadAsync``` call ```Read``` synchrounously and then returning the result.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Encryption/ZipAESStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Encryption/ZipAESStream.cs
@@ -1,6 +1,8 @@
 using System;
 using System.IO;
 using System.Security.Cryptography;
+using System.Threading;
+using System.Threading.Tasks;
 using ICSharpCode.SharpZipLib.Core;
 
 namespace ICSharpCode.SharpZipLib.Encryption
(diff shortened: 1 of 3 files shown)


Numpsy:
>>> just making ReadAsync call Read synchrounously and then returning the result

which would be the fastest change if we want a fix ASAP, but not as nice as a real async impl, which fits in more with the other async supporting changes.

src/ICSharpCode.SharpZipLib/Encryption/ZipAESStream.cs
```diff
@@ -105,68 +121,127 @@ private int ReadAndTransform(byte[] buffer, int offset, int count)
...
+					throw new Exception("Internal error missed auth code"); // Coding bug
+																			// Final block done. Check Auth code.
+				byte[] calcAuthCode = _transform.GetAuthCode();
+				for (int i = 0; i < AUTH_CODE_LENGTH; i++)
+				{
+					if (calcAuthCode[i] != _slideBuffer[_slideBufStartPos + i])
 					{
-						if (calcAuthCode[i] != _slideBuffer[_slideBufStartPos + i])
-						{
-							throw new Exception("AES Authentication Code does not match. This is a super-CRC check on the data in the file after compression and encryption. \r\n"
```
Numpsy, line 147: On a related note, should this be throwing ZipException rather than Exception?

PR #578, opened 2021-02-14
Fix typos in the StreamDecodingException doc comments

Numpsy:
Just something I noticed while looking at something else.


piksel (approved):
Yeah, these ones are quite common

PR #579, opened 2021-02-14
Implement ReadAsync in ZipAESStream, extra simple version

Numpsy:
Possible extra trivial version of #576, just for testing and a potential quick fix (whereby it doesn't attempt to actually be async)


Numpsy:
@piksel Any thoughts on applying the simple approach as a quick fix for for #572, given that I don't *think* a 'full' async version would actually do async reads at this point unless PartialInputStream was made async as well  (ref #589)?

(It'd be nicer to make everything more fully async, not sure if there's any  thoughts of a timespan for a next maintenance release vs. doign that though)

piksel:
Yeah, I agree.

Issue #44, opened 2013-12-29
Calling Close or Dispose on zip entry stream, and finalizers


dgrunwald:
[SD-1880](http://bugtracker.sharpdevelop.net/issue/ViewIssue.aspx?id=1880), originally created on 1/27/2012 05:55:20 by David Pierson

Several posters have reported issues when calling Flush Close and
Dispose on the stream they get from GetInputStream. 

 Worth remarking for any readers, it is recommended not to call any of
those methods against the entry stream. 

 Threads:
 [System.ObjectDisposedException: Cannot access a closed
file](http://community.sharpdevelop.net/forums/t/7335.aspx)

[http://community.sharpdevelop.net/forums/t/14397.aspx](http://community.sharpdevelop.net/forums/t/14397.aspx)

 Even without calling Close/Dispose the .NET GC may finalise the outer
stream while still using the entry stream. Believe this has been fixed
in 0.85.5 - worth desk checking.
 Thread:
 [Random bug while
unzipping](http://community.sharpdevelop.net/forums/t/6881.aspx)

 "There is a flaw in the dispose method of ZipFile also which isnt
checking for disposing before closing streams. The f
 inalizer for ZipFile can indeed be removed as Daniel states.  There is
nothing for it to do once you correct the flaw above."

PR #589, opened 2021-03-01
[WipAdd an implementation of ReadAsync to PartialInputStream

Numpsy:

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -9,6 +9,8 @@ using System.Collections.Generic;
 using System.IO;
 using System.Security.Cryptography;
 using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace ICSharpCode.SharpZipLib.Zip
 {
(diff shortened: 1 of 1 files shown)


Numpsy:
@piksel Do you have any thoughts on the importance of the locks used in the existing PartialInputStream read functions, or on the best way to approach them in any async implementation?

The common suggestion seems to be to use ```SemaphoreSlim``` instead of ```lock()``` in async functions which would be doable, but then you might need to change ZipFile itself to manage the semaphore, and as ZipFile.DisposeInternal tries to lock the ```baseStream``` itself that might have issues of its own (though whether trying to lock on and dispose of a member variable in a function thats called from a finalizer is sensible is another question, refs #44  )

piksel:
I think using [Stream.Synchronized(Stream)](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.synchronized?view=net-5.0) should be roughly equivalent to using `SemaphoreSlim` (preventing _our_ code from altering the stream). 

Now, the problem is that it would still not be restricting code _outside_ our library from altering the base stream. At least I don't think so. Using `lock` should prevent all access (since the lock is on the reference to the stream itself), but there is a lot of guesswork going on here. 😓

Numpsy:
As in, create the partial input stream from a synchronized stream, rather than the base stream?

Would that work in the case of ```PartialInputStream.Read``` which does a seek on the basestream followed by a read? (as in, would that be equivalent to locking ```Seek``` and ```Read``` independantly rather than both as a unit, such that you could potentially allow more than one thread to do a Seek while others are pending read?)

Numpsy:
I don't know if you can *really* prevent a caller who constructed a ZipFile from an arbitrary stream from changing that stream themselves outside of the library code - even if ```lock``` did prevent that, it's only locked part of the time anyway, so there would still be times when it could change

piksel:
Yeah, I just don't know what else this is meant to do? The original code is synchronous, so what was the point?

Also `Stream.Synchronized` doesn't help that much as it's just synchronizing calls to read/write etc. Replacing the lock with a `SemaphoreSlim` is probably the sanest thing to do.

Numpsy:
> Yeah, I just don't know what else this is meant to do? The original code is synchronous, so what was the point?

I guessed that it was to try to avoid issues if the user tries to simultaneously read the content from multiple zip entries at the same time from different threads (given that nothing in the library itself should try to do anything like that).

The lock at https://github.com/icsharpcode/SharpZipLib/blob/5eea92acf228c72af124aa2fd7ca1b36987cdbbe/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs#L3354 looks more odd to me - not sure if that would be doing anything other than trying to protect a caller from disposing the ZipFile while they're still reading from it?

PR #592, opened 2021-03-07
feat(zip): better string encoding handling

piksel:
This replaces the global static `ZipStrings` singleton with instances of `StringCodec`, which will:
- Remove encoding configuration from a shared global state
- Allow for different defaults for input and output
- Explicitly override the encodings used for `ZipCrypto` and zip archive comments (the one in the _Central Directory_, not the individual entry comments).
- Use "Unicode" for new entries (unless overriden)
- Make it much more clear (hopefully) how and why different encodings are used.

Since the only way to customize Encoding handling in zipfiles until now has been to modify the static `ZipStrings` singleton, and we don't want to break backwards compatibility, it will now use separate instances for all `ZipFile/Zip*Stream` instances **unless** `ZipStrings` has been modified.
It's still not a great API, but the properties are marked as deprecated and should be removable in a later release.


piksel:
...oh well...
image

Numpsy:
> 
> 
> ...oh well...
> image

Is there actually a reason for the test project to use a custom nuget.config?
(I have an idea that MS have moved some of the things they used to host there elsewhere anyway?)

piksel:
I didn't even realize we had one.  Removed it, and half of the restores work now... I guess I'll give it some time. 🤷‍♀️

Numpsy:
removed from the unit tests and test bootstrapper in #594

piksel:
That seems to have fixed it. Now there is only this... 😁
image

src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs
```diff
@@ -20,175 +25,100 @@ static ZipStrings()
...
-				{
-					throw new ArgumentOutOfRangeException(nameof(value));
-				}
+		public bool ForceZipLegacyEncoding { get; set; }
 
-				codePage = value;
-			}
-		}
+		/// <summary>
+		/// The default encoding used for ZipCrypto passwords in zip files, set to <see cref="SystemDefaultEncoding"/>
```
Numpsy, line 38: Is that something that might differ between .NET 4.x and .NETCore on Windows? (might be UTF-8 on Core, something else on 4.8?)

src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs
```diff
@@ -20,175 +25,100 @@ static ZipStrings()
...
-		/// </param>
-		/// <returns>Converted array</returns>
-		public static byte[] ConvertToArray(int flags, string str)
-			=> (string.IsNullOrEmpty(str))
-				? Empty.Array<byte>()
-				: EncodingFromFlag(flags).GetBytes(str);
+		public Encoding ZipCryptoEncoding
+		{
+			get => _zipArchiveCommentEncoding ?? DefaultZipCryptoEncoding;
+			set => _zipArchiveCommentEncoding = value;
```
Numpsy, line 121: Is setting _zipArchiveCommentEncoding correct? (or if the intention is to share an encoding between ZipCrypto and ZipArchiveComment, maybe the backing field could have a more explanitory name?)

piksel: Nope, thats a typo.

src/ICSharpCode.SharpZipLib/Zip/ZipStrings.cs
piksel: Yeah, https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.default?view=net-5.0#the-default-property-on-net-core 

Does it matter though? It still seems like the best defaults for non-windows and core-on-windows consumers... They can override the encoding if they know what to use, but we cant detect the OS defaults without interop afaik.

Numpsy: >Does it matter though?
Quite possibly not (or possibly it's the only thing to use that doesn't need the extra encoding packages registered anyway).

What's the situation with using Encoding.Default vs Encoding.GetEncoding(0) on core? (i.e. if the behaviour of GetEncoding(0) might vary depending on which providers have been registered, and .Default is constant, which one is best?)

piksel: Yeah, if the consumer registers additional providers *before* the **static** constructor of `ZipStrings` it will return another value. It's probably better to use `Encoding.Default` instead, and on windows, explicitly call [GetACP](https://docs.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getacp).

PR #593, opened 2021-03-07
Simplify DropPathRoot and fix out of bounds issue

### LINKED ISSUE #56: Bug in DropPathRoot - IndexOutOfBoundsException
There is a bug in DropPathRoot (in WindowsPathUtils.cs / ICSharpCode.SharpZipLib.Core.WindowsPathUtils) that causes an IndexOutOfBoundsException:

This can be seen for paths such as "\\server".  It works correctly (as it is) for paths such as "\\server\share".

is:

``` c#
// Scan for two separate elements \\machine\share\restofpath
while ((index <= path.Length) &&
    (((path[index] != '\\') && (path[index] != '/')) || (--elements > 0))) {
    index++;
}
```

should likely be (but I think other modifications will be needed as well to obtain desired behavior):

``` c#
// Scan for two separate elements \\machine\share\restofpath
while ((index < path.Length) &&
    (((path[index] != '\\') && (path[index] != '/')) || (--elements > 0))) {
    index++;
}
```

Also, since "index" starts at 2, a few lines above, I'm not sure what the purpose of "elements" is -- possibly a left over?

Thoughts?

piksel:
Fixes #56

This did grow a bit, since a lot of stuff was discovered while working on this.
Prior to this, all tests (and indeed the behaviour) for handling paths just presumed that the platform was Windows.
The core problem method (DropPathRoot) has now been changed to use the framework/platforms path handling methods, but with some special exceptions for .NET < 4.6.2.
This does feel a bit hackish, but I think it's better than the previous method, which was to implement custom code for determining what a root path is. If that code fails to notice something that the OS would consider a absolute path it could lead to vulnerabilities. If this implementation fails to notice it, it would instead throw, which could then be fixed.
And also, since this only happens in older frameworks, in most cases it would just work as intended.

If creating from a windows OS, it will strip what the OS considers to be a root path from the file name, which ensures "correct" entries in our archive, and on reading entries from an archive, it will strip them as appropriate for the OS.

This PR also includes updates to tests which now have separated versions for windows and non-windows platforms, and also pipes the output of `7z` to the test context, so that it will only be displayed if a test fails.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Core/PathUtils.cs
+++ b/src/ICSharpCode.SharpZipLib/Core/PathUtils.cs
@@ -1,4 +1,6 @@
+using System;
 using System.IO;
+using System.Linq;
 
 namespace ICSharpCode.SharpZipLib.Core
 {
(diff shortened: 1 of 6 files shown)


src/ICSharpCode.SharpZipLib/Core/PathUtils.cs
```diff
@@ -15,48 +15,9 @@ public static class PathUtils
 		/// <remarks>Unlike the <see cref="System.IO.Path"/> class the path isn't otherwise checked for validity.</remarks>
```
Numpsy, line 15: Referring to the comment about not checking for validity - I can't recall if this is ever called on reading a file or just creating? - are there any situations where GetPathRoot might throw on .NET 4.x if it thinks something is invalid (e.g. same sort of situation as #341)

piksel: Yeah, you are probably right. I'll add a test for this and fix it if it triggers anything.
If it does, it is probably better to use another code path for cleaning "input" file names, rather than sharing it with the "output" ones.

piksel: The results are in!
image
The usual reserved characters for file names in windows are `\ / : * ? " < > |`, but out of those, only `< > | "` are listed in `Path.GetInvalidPathChars()` and throws on net46.
I guess the easiest way to solve it is to just replace those chars with `_`?

piksel: Added an optional `replaceInvalidChars` to the arguments as well, since the work is already done, but per default it returns the stripped path, **including** any invalid path characters.

Numpsy: I've forgotten all the details offhand about how DropPathRoot interacts with the other name transform/cleaning stuff, but looking at the comment about long paths reminded me of the ```\\?\``` long path syntax on Windows, the handling of which might be another reason to leave it up to the built in functions instead of doing things manually

piksel: It's used in `WindowsNameTransform` (for zip -> windows file name), `PathNameTransform` and `ZipFileTransform` for `ZipOutputStream` and `FastZip`/`ZipEntryFactory` respectively.

Issue #590, opened 2021-03-03
GZipOutputStream.Password doesn't work


proy27:
### Expected behavior
GZipOutputStream.Password doesn't work

### Actual behavior
Tell us what happens instead

### Version of SharpZipLib
1.3.1
### Obtained from (only keep the relevant lines)
- Package installed using NuGet

piksel:
Please fill out the issue template. You are currently claiming that you don't expect it to work (whatever that even means).

The GZip format does not support encryption, so I am not even sure what you are trying to do...

Numpsy:
It *does* seem a bit unfortunate that ```DeflaterOutputStream``` has a public ```Password``` property, given that the encryption it refers to only actually applies to zip files (and the doc comments on the property don't say anything to that effect, so that might be misleading?)

Numpsy:
Fixed by #604 ?

PR #604, opened 2021-03-14
Move the Password property from DeflaterOutputStream into ZipOutputStream

Numpsy:
refs #590 

This would be a breaking API change, but - is there actually a need for the Password property to be on DeflaterOutputStream, given that it only applies to creating encrypted Zip files? (the property isn't actually referenced by DeflaterOutputStream anyway, only from ZipOutputStream)

Related:
The ```InitializePassword``` and ```InitializeAESPassword``` functions in DeflaterOutputStream only seem to be called from ZipOutputStream as well, though they set the (presently private) ```cryptoTransform_``` member as well, so moving them as well would need a further change.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/DeflaterOutputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/DeflaterOutputStream.cs
@@ -153,37 +153,12 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 
 		#region Encryption
 
-		private string password;
-
 		private ICryptoTransform cryptoTransform_;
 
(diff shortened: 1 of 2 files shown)


Numpsy:
@piksel would you consider the public API change as acceptable, before I see about doing anything else with this?

piksel:
Yeah, I think it's good idea. It's a breaking change on a technical level, but if there is any code out there relying on it, it's probably not doing what the author intended anyway (as shown by the referenced issue).

Numpsy:
hmm, moved the code that sets up the crypto transforms and didn't notice till looking at the coverage results that the existing code initialises the static ```_aesRnd``` member where it's declared and also checks it for null and tries to set it up where it's used, can have a look at tidying that up as well.

PR #618, opened 2021-05-03
Add CLSCompliant Attribute

tiesnes:
From [Microsoft](https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components):
> If your component conforms to the Common Language Specification, it is guaranteed to be CLS-compliant and can be accessed from code in assemblies written in any programming language that supports the CLS. You can determine whether your component conforms to the Common Language Specification at compile time by applying the CLSCompliantAttribute attribute to your source code.

The attribute used to be part of the project. It got removed with be53259f188e39c0965dfe8d36448ddbd706403b.


piksel (approved):

piksel:
I think adding it back to the assembly would be fine, as I don't think there are any non-CLS features being exposed, but what is this trying to solve?

tiesnes:
Adding the attribue allows apps written in other languages (F# or VB) to consume the assembly.

tiesnes:
Is there any chance this will be included in the v1.4 version?

piksel:
Yeah, sorry. What is it you are trying to solve with this? I know what the attribute does, but it's not needed for VB and F# usage afaik. Do you have a specific problem?

tiesnes:
We do have a specific problem: We use SharpZipLib in one of our legacy libraries. Unfortunately, it exposes a few classes of SharpZipLib through its public interface.
Our library and its consumers are all CLS-compliant. SharpZipLib was compliant too, until the update from 0.86.0 to 1.3.1. Since then, every consumer of the legacy library gets compiler warnings for the exposed, non-compliant classes.
Surely, we could solve our problem by changing the public interface and hiding the non-compliant classes behind a new level of abstraction. But these changes come with risks, whereas reintroducing the attribute is a low-risk operation.

piksel:
Okay, this should be fine then. Could you fix the merge conflict? I cannot do it through Github since it will directly commit to master.
```cs
using System;
using System.Runtime.CompilerServices;

[assembly: CLSCompliant(true)]
[assembly: InternalsVisibleTo("ICSharpCode.SharpZipLib.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b9a14ea8fc9d7599e0e82a1292a23103f0210e2f928a0f466963af23fffadba59dcc8c9e26ecd114d7c0b4179e4bc93b1656b7ee2d4a67dd7c1992653e0d9cc534f7914b6f583b022e0a7aa8a430f407932f9a6806f0fc64d61e78d5ae01aa8f8233196719d44da2c50a2d1cfa3f7abb7487b3567a4f0456aa6667154c6749b1")]
```

tiesnes:
Thank you, the merge conflict is resolved.

Before accepting this pull request, have a look at the interface `ITaggedData` in ZipExtraData.cs.
The interface is not CLS-compliant, it declares a property of type `ushort`.
There will be compiler warnings because of this.

How would you like to proceed?

PR #619, opened 2021-05-06
[WIP] experiment - Use SemaphoreSlim to prevent concurrent basestream access

Numpsy:
A spin off / follow up to #589 which tries to use ```SemaphoreSlim``` to prevent concurrent access to the base stream instead of ```lock()```

Another followup thought after looking at the bits that lock a bit more:

The ```TestArchive``` function:
1) calls ```TestLocalHeader```, which locks the stream for the duration of the operation
2) reads the stream content via ```GetInputStream```, which returns the locking PartialInputStream
3) uses ```ZipHelperStream``` to read the data descriptor, which I don't think does any locking

so might (3) be a hole where multiple things could read the stream at once?

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -9,6 +9,8 @@ using System.Collections.Generic;
 using System.IO;
 using System.Security.Cryptography;
 using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace ICSharpCode.SharpZipLib.Zip
 {
@@ -1079,8 +1081,10 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// <returns>The offset of the entries data in the file</returns>
 		private long TestLocalHeader(ZipEntry entry, HeaderTest tests)
 		{
-			lock (baseStream_)
+			try
 			{
+				this.baseStreamSemaphore.Wait();
+
 				bool testHeader = (tests & HeaderTest.Header) != 0;
 				bool testData = (tests & HeaderTest.Extract) != 0;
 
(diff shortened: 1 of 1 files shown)


src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -3349,11 +3357,21 @@ private void DisposeInternal(bool disposing)
...
 				entries_ = Empty.Array<ZipEntry>();
 
-				if (IsStreamOwner && (baseStream_ != null))
+				if (disposing)
 				{
-					lock (baseStream_)
+					try
 					{
-						baseStream_.Dispose();
+						this.baseStreamSemaphore.Wait();
```
Numpsy, line 3364: I did this the other day and wasn't sure about it, and it still feels off - if the idea of the existing lock is to prevent the stream being disposed while being read (rather than 'just' being disposed in between reads) then that needs to be maintained, but then you might also need to avoid disposing the semaphore while another thread is using it, which just makes things more complicated.

Still not sure if the lock is doing anything other than trying to save users from their own mistakes though (preventing an explicit attempt to dispose the archive while still using it) :-(

Numpsy:
I've read elsewhere that you don't *need* to dispose the semaphore if you aren't using it's wait handle property directly, but I don't know if that's something to depend on.

Extra related question: Should the Async stream calls use ```ConfigureAwait``` ?

piksel:
Regarding `ConfigureAwait`:
> if you’re writing general-purpose library code, use ConfigureAwait(false)

from https://devblogs.microsoft.com/dotnet/configureawait-faq/#when-should-i-use-configureawaitfalse

src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
piksel: Yeah, I'm not even sure that that is something we should try to manage. Tried looking up when the locking was added, but it lead me to [this wonderful commit](https://github.com/icsharpcode/SharpZipLib/commit/88e49b8a6acd09c647809351bf2679920582a429) with 124,411 additions and 10,927 deletions... Great.

Numpsy:
Yes, just thought i'd check as it can mean adding it in many places, and #574 didn't seem to do it.

piksel:
No, but that's just my bad 😅

PR #627, opened 2021-05-08
fix the expected/actual value ordering in a few unit test asserts

Numpsy:
rather trivial, but still


Numpsy:
Not sure what's going on with the coverage results - this shouldn't effect coverage at all, even less so in BZip2OutputStream

piksel:
No, there is something really odd going on. I have tried fixing it by reducing the amount of test coverage results being uploaded, but it's still bonkers.

PR #633, opened 2021-05-11
Suppress some localization related code analyzer suggestions

Numpsy:
Spin off from #449 that disables some of the localization related analyzers, to decide if disabling the set is ok (for now at least)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -37,3 +37,18 @@ end_of_line = crlf
 [*.yml]
 indent_style = space
 indent_size = 2
+
+# C# source
+[*.{cs}]
+
+# CA1303: Do not pass literals as localized parameters
+dotnet_diagnostic.CA1303.severity = none
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -37,3 +37,18 @@ end_of_line = crlf
...
+dotnet_diagnostic.CA1303.severity = none
+
+# CA1305: Specify IFormatProvider
+dotnet_diagnostic.CA1305.severity = none
+
+# CA1304: Specify CultureInfo
+dotnet_diagnostic.CA1304.severity = none
+
+# CA1307: Specify StringComparison for clarity
+dotnet_diagnostic.CA1307.severity = none
```
piksel, line 54: I'm not too sure about these... They might still lead to some unwanted behaviour when running in various non-english environments...
It would probably be better to actually fix these.

PR #611, opened 2021-04-18
Bzip input stream simple vectorization

konrad-kruczynski:
Hi guys.
Here's the simple approach to get some speed-up on BZip2 decompression. Rotation loop (`yy[j] = yy[j -1]`) is here vectorized automatically, i.e. by means of `Vector<byte>` instead of e.g. `Vector128<byte>` tied to a specific platform. The API is available from the .NET Core/Standard 2.1, however the real gain starts in .NET Core 3.

First commits add .NET Core 3.1 as a target platform for performance tests and such a test for BZip2 decompression. The last commit is the actual vectorization. Here are some results from two Intel machines (one on Windows and one on a rather antique MacBook Air).

---
**Without** vectorization:
First machine
```
|         Method |        Job |     Toolchain |    Mean |    Error |   StdDev | Ratio |
|--------------- |----------- |-------------- |--------:|---------:|---------:|------:|
| DecompressData | Job-USQLOW | .NET Core 2.1 | 4.595 s | 0.0185 s | 0.0164 s |  0.97 |
| DecompressData | Job-NMSPHA | .NET Core 3.1 | 4.730 s | 0.0490 s | 0.0434 s |  1.00 |
| DecompressData | Job-ORXZIN |        net461 | 4.723 s | 0.0314 s | 0.0279 s |  1.00 |
```
Second machine
```
|         Method |        Job |     Toolchain |    Mean |    Error |   StdDev |
|--------------- |----------- |-------------- |--------:|---------:|---------:|
| DecompressData | Job-TBYXJE | .NET Core 2.1 | 8.177 s | 0.0607 s | 0.0538 s |
| DecompressData | Job-ZFNFSZ | .NET Core 3.1 | 8.343 s | 0.1529 s | 0.2717 s |
```
----
**With** vectorization:
First machine
```
|         Method |        Job |     Toolchain |    Mean |    Error |   StdDev | Ratio |
|--------------- |----------- |-------------- |--------:|---------:|---------:|------:|
| DecompressData | Job-JSFVZF | .NET Core 2.1 | 4.534 s | 0.0386 s | 0.0302 s |  0.99 |
| DecompressData | Job-TNNLLN | .NET Core 3.1 | 2.335 s | 0.0252 s | 0.0211 s |  0.51 |
| DecompressData | Job-THQXFG |        net461 | 4.590 s | 0.0217 s | 0.0181 s |  1.00 |
```
Second machine
```
|         Method |        Job |     Toolchain |    Mean |    Error |   StdDev |
[...]

PR #635, opened 2021-05-13
Fixed mismatched framework directives for vectorized memory move

modio-jackson:


piksel:
This came up code review, https://github.com/icsharpcode/SharpZipLib/pull/611#discussion_r623915992
Do you disagree?

src/ICSharpCode.SharpZipLib/BZip2/BZip2InputStream.cs
```diff
@@ -1,3 +1,7 @@
+#if !NETSTANDARD2_0 && !NETFRAMEWORK
```
piksel, line 1: This effectively does the same thing, even if we are not targeting any .NET core ATM, it would still apply if we (or someone compiling from source) did.
Not sure that this is clearer, it explicitly states what versions are supported, instead of which ones isn't...
```suggestion
#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP3_1_OR_GREATER
```

modio-jackson:
Unity's build engine defines `NET_STANDARD_2_0` rather than `NETSTANDARD2_0` which causes the library to fail to compile. (https://docs.unity3d.com/Manual/PlatformDependentCompilation.html)
`!NETSTANDARD2_0 && !NETFRAMEWORK` resolves to `true`, and contains code requiring the definition of `VectorSize`.
`NETSTANDARD2_1` resolves to `false`, causing `VectorSize` to remain undefined.

modio-jackson:
> 
> 
> Unity's build engine defines `NET_STANDARD_2_0` rather than `NETSTANDARD2_0` which causes the library to fail to compile. (https://docs.unity3d.com/Manual/PlatformDependentCompilation.html)
> `!NETSTANDARD2_0 && !NETFRAMEWORK` resolves to `true`, and contains code requiring the definition of `VectorSize`.
> `NETSTANDARD2_1` resolves to `false`, causing `VectorSize` to remain undefined.

konrad-kruczynski:
Frankly speaking I didn't know that `_OR_GREATER` were available. In that case I agree that this is much better. Moreover, this will also work for .NET 5.

piksel (approved):

PR #642, opened 2021-06-07
fix(zip): handle iterating updated entries in ZipInputStream

### LINKED ISSUE #639: ZipInputStream cannot read in-place updated archives
### Steps to reproduce
1.unity2018 empty project
2.build apk by Internal build system
3.unzip the apk
code:
```
ZipInputStream s = new ZipInputStream(File.OpenRead(file));	
ZipEntry theEntry;
while ((theEntry = s.GetNextEntry()) != null)
```
test apk file
[https://pan.baidu.com/s/1UJLdYwWMvpuoVjqCZaB-Hg](url)
net disk key: 5ygk

### Expected behavior
unzip the apk success

### Actual behavior
we get Incorrect and Garbled string and then Run to GetNextEntry while has some exception as "Wrong Local header signature: 0x50000000"

### Version of SharpZipLib
v1.3.2

source code from github

piksel:
Currently `ZipInputStream` expects all archive entries to immediately follow the previous one. This is normally the case for archives created in one go, but necessarily true when they have been updated.
This PR allows `ZipInputStream` to skip any padding between entries and consequently read "direct-updated" archives.

Fixes #639 

Curiously, 7-zip does not seem to handle these files either and just show all entries found before the first "padded" one. And of course, when updating it uses the equivalent of `ZipFile`s `FileUpdateMode.Safe` to avoid the issue.


test/ICSharpCode.SharpZipLib.Tests/Zip/StreamHandling.cs
```diff
@@ -556,5 +564,43 @@ public void ShouldThrowDescriptiveExceptionOnUncompressedDescriptorEntry()
...
+		{
+			using (var tempFile = new Utils.TempFile())
+			{
+				var fi = new FileInfo(tempFile.Filename);
+				
+				using (var zf = ZipFile.Create(tempFile.Filename))
+				{
+					zf.BeginUpdate();
+					// Add a "large" file
+					zf.Add(new MemoryDataSource(new byte[1024]), "FirstFile", CompressionMethod.Stored);
```
piksel, line 580: Currently only tests skipping null-bytes (since the first file content is used an array of `default(byte)` (`0`).
Add some non-zero data to the end of the data source to additionally test non-null seeking:
https://github.com/icsharpcode/SharpZipLib/blob/c6ccef30ed603897e665d4d65afae73b257dfb70/src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs#L324-L328

PR #667, opened 2021-09-20
Update TarInputStream.cs

Svisstack:


piksel:
Why not use https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copytoasync?view=net-5.0#System_IO_Stream_CopyToAsync_System_IO_Stream_ ?

Svisstack:
True, but the non-async version should not exist and it's misleading at first sight.

piksel:
Yeah, I didn't even know we had one. We should probably deprecate it...

PR #670, opened 2021-10-07
Remove the test bootstrapper

Numpsy:
As suggested in #630 


piksel:
Yeah, there are still randomness in the test cases which causes some areas to only be covered with the "correct" seed.

piksel:
I have gone through all tests using `Random` and replaced them with deterministic pseudo random data in #671. This *should* stabilize the test coverage...

PR #678, opened 2021-10-22
feat(fastzip): allow custom filesystemscanner

### LINKED ISSUE #676: FastZip.CreateZip - how to deal with buggy Directory enumeration on .net 5
Not really a bug, more like a feature suggestion.

### Steps to reproduce
1. In .net 5, directory enumeration has bugs with WebDav (https://github.com/dotnet/runtime/issues/46723#issuecomment-928090089), not fixed in .net 6 and with no other immediate official solution in sight;
2. This causes crashes when calling `FastZip.CreateZip` as the GetDirectories/GetFiles return invalid data;
3. A workaround is available - use something like this on the GetDirectories-returned data:

```
public static string[] FixupNet5Enumeration(string[] items)
{
    return items
        .Select(t => t.TrimEnd('\0'))
        .Where(t => {
            var itemName = Path.GetFileName(t);
            return itemName != "." && itemName != "..";
        })
        .ToArray();
}
```
4. SharpZipLib nicely wraps the `Directory.GetDirectories` call into the `FileSystemScanner` class. However it is not possible to supply custom `FileSystemScanner`s to `FastZip.CreateZip`. It would be helpful in this case as I could just copy that class and insert the required fixup calls. It would be necessary to create an interface around `FileSystemScanner` and to change to `public` the method 

https://github.com/icsharpcode/SharpZipLib/blob/612969e574fd9d922314f24923792e343871f102/src/ICSharpCode.SharpZipLib/Zip/FastZip.cs#L477

What do you think?

### Expected behavior
`FastZip.CreateZip` should offer an override to specify custom `FileSystemScanner`s.

### Actual behavior
`FastZip.CreateZip` throws exception as the paths returned by `Directory.GetDirectories` are not valid paths.

### Version of SharpZipLib
1.3.3

piksel:
Fixes #676 

Or rather, allows it to be handled with a derived class:

```cs
  new FastZip().CreateZip(outStream, srcDir, recurse: true, new CustomFileSystemScanner(filter), leaveOpen: true);
```

```cs
class CustomFileSystemScanner:  FileSystemScanner 
{
	static IEnumerable<string> FixupNet5Enumeration(IEnumerable<string> items)
		=> items
			.Select(t => t.TrimEnd('\0'))
			.Where(t => {
				var itemName = Path.GetFileName(t);
				return itemName != "." && itemName != "..";
			});

	protected override IEnumerable<string> GetDirectories(string directory) 
		=> FixupNet5Enumeration(base.GetDirectories(directory));

	protected override IEnumerable<string> GetFiles(string directory) 
		=> FixupNet5Enumeration(base.GetFiles(directory));

	public CustomFileSystemScanner(string ff): base(ff) {}
	public CustomFileSystemScanner(string ff, string df): base(ff, df) {}
	public CustomFileSystemScanner(IScanFilter ff): base(ff){}
	public CustomFileSystemScanner(IScanFilter ff, IScanFilter df) : base(ff, df) {}
}
```

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Core/FileSystemScanner.cs
+++ b/src/ICSharpCode.SharpZipLib/Core/FileSystemScanner.cs
@@ -1,4 +1,7 @@
 using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 
 namespace ICSharpCode.SharpZipLib.Core
 {
@@ -9,42 +12,24 @@ namespace ICSharpCode.SharpZipLib.Core
 	/// </summary>
 	public class ScanEventArgs : EventArgs
 	{
-		#region Constructors
-
 		/// <summary>
 		/// Initialise a new instance of <see cref="ScanEventArgs"/>
 		/// </summary>
 		/// <param name="name">The file or directory name.</param>
 		public ScanEventArgs(string name)
 		{
-			name_ = name;
+			Name = name;
 		}
-
-		#endregion Constructors
 
 		/// <summary>
 		/// The file or directory name for this event.
 		/// </summary>
-		public string Name
-		{
-			get { return name_; }
-		}
+		public string Name { get; }
 
 		/// <summary>
 		/// Get set a value indicating if scanning should continue or not.
 		/// </summary>
-		public bool ContinueRunning
-		{
-			get { return continueRunning_; }
-			set { continueRunning_ = value; }
-		}
-
-		#region Instance Fields
-
-		private string name_;
-		private bool continueRunning_ = true;
-
-		#endregion Instance Fields
+		public bool ContinueRunning { get; set; } = true;
 	}
 
 	/// <summary>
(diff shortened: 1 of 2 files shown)


src/ICSharpCode.SharpZipLib/Core/FileSystemScanner.cs
```diff
@@ -496,31 +389,29 @@ private void ScanDir(string directory, bool recurse)
...
 			}
 
-			if (alive_ && recurse)
+			if (!alive_ || !recurse) return;
+			
+			try
 			{
-				try
+				var subDirectories = GetDirectories(directory)
+					.Where(d => d != null && directoryFilter_.IsMatch(d));
```
piksel, line 397: Check `direcotryFilter_` for null. Perhaps by
```suggestion
				Func<bool, string> filter = directoryFilter_ == null 
					? (d) => true 
					: (d) =>  directoryFilter_.IsMatch(d);
				var subDirectories = GetDirectories(directory)
					.Where(d => d != null && filter(d));
```

piksel (commented):
Needs more tests, but otherwise looks OK.

PR #687, opened 2021-11-18
Fix NullReferenceException in PathUtils.DropPathRoot if Path.GetPathR…

dime-makarov:
The Path.GetPathRoot method is able to return null.
PathUtils.DropPathRoot should handle this case.


lmponte:
This has affected our project, we were using this library to zip and unzip a simple message to be stored elsewhere, and passing the `ZipEntry` name as `string.Empty`. After upgrading from v1.3.1 to v1.4.2, this started throwing an exception.

piksel:
Yeah, this seems like it should obviously be handled. I want add a test case for this, but otherwise this should be mergable.

piksel:
Since the source branch is removed, this is superseded by #828.

PR #746, opened 2022-04-10
feat(tar): support for async streams

nathan-c:

Relates to #223 

I have taken a first pass at making TAR support async streams which will help with streaming to and from network streams. I have also tried to dramatically reduce memory allocations associated with reading from and writing to TAR's.

I haven't added new tests yet or really checked what the coverage is like but if you are happy with how it looks as it, I can do that.

Below are some benchmarks run before and after my changes:  
Before:  
|                  Method |        Job |     Toolchain |       Mean |    Error |   StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------------ |----------- |-------------- |-----------:|---------:|---------:|--------:|-------:|------:|----------:|
|      ReadTarInputStream | Job-IFGTKW | .NET Core 3.1 |   122.3 μs |  1.71 μs |  1.43 μs | 33.2031 | 1.2207 |     - | 1112144 B |
| ReadTarInputStreamAsync | Job-IFGTKW | .NET Core 3.1 | 1,081.9 μs | 21.64 μs | 58.50 μs | 37.1094 | 0.9766 |     - | 1227248 B |
|      WriteTarOutputStream | Job-IFGTKW | .NET Core 3.1 |    50.54 μs |  0.339 μs |  0.283 μs | 0.3052 |     - |     - |   12008 B |
| WriteTarOutputStreamAsync | Job-IFGTKW | .NET Core 3.1 | 1,135.09 μs | 22.460 μs | 51.154 μs | 3.9063 |     - |     - |  126985 B |

After:  
|                  Method |        Job |     Toolchain |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------ |----------- |-------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
|      ReadTarInputStream | Job-KKAMPY | .NET Core 3.1 | 221.3 μs | 0.76 μs | 0.64 μs |     - |     - |     - |     952 B |
| ReadTarInputStreamAsync | Job-KKAMPY | .NET Core 3.1 | 227.1 μs | 4.05 μs | 3.79 μs |     - |     - |     - |    1096 B |
|      WriteTarOutputStream | Job-KKAMPY | .NET Core 3.1 | 149.7 μs | 2.71 μs | 2.54 μs |     - |     - |     - |     449 B |
| WriteTarOutputStreamAsync | Job-KKAMPY | .NET Core 3.1 | 146.7 μs | 0.67 μs | 0.56 μs |     - |     - |     - |     448 B |

### DIFF
--- a/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
+++ b/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -36,10 +36,12 @@ Please see https://github.com/icsharpcode/SharpZipLib/wiki/Release-1.3.3 for mor
 
 	<ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
 		<PackageReference Include="System.Memory" Version="4.5.4" />
+		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.2" />
 	</ItemGroup>
 
 	<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
 		<PackageReference Include="System.Memory" Version="4.5.4" />
+		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.2" />
 	</ItemGroup>
 
 	<ItemGroup>
--- a/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
+++ b/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
@@ -33,8 +33,16 @@ Please see https://github.com/icsharpcode/SharpZipLib/wiki/Release-1.3.3 for mor
 	<ItemGroup>
 		<PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
 	</ItemGroup>
-	 
-	 <ItemGroup>
+
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
+		<PackageReference Include="System.Memory" Version="4.5.4" />
+	</ItemGroup>
+
+	<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+		<PackageReference Include="System.Memory" Version="4.5.4" />
+	</ItemGroup>
+
+	<ItemGroup>
 	   <None Include="../../assets/sharpziplib-nuget-256x256.png">
       <Pack>True</Pack>
       <PackagePath>images</PackagePath>
--- a/src/ICSharpCode.SharpZipLib/Tar/TarBuffer.cs
+++ b/src/ICSharpCode.SharpZipLib/Tar/TarBuffer.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Buffers;
 using System.IO;
 
 namespace ICSharpCode.SharpZipLib.Tar
@@ -207,7 +208,7 @@ namespace ICSharpCode.SharpZipLib.Tar
 		{
 			blockFactor = archiveBlockFactor;
 			recordSize = archiveBlockFactor * BlockSize;
-			recordBuffer = new byte[RecordSize];
+			recordBuffer = ArrayPool<byte>.Shared.Rent(RecordSize);
 
 			if (inputStream != null)
 			{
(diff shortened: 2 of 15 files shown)


piksel:
Wow, for some reason I thought `Stream.WriteAsync` was not available in .NET 4.5... 
This should make the `async` versions of `Zip[In/Out]putStream` much clearer.

Anyway, this mostly looks great, but it does lower the test coverage a bit.

src/ICSharpCode.SharpZipLib/Tar/TarBuffer.cs
```diff
@@ -550,36 +606,77 @@ private void WriteFinalRecord()
...
 
-			outputStream.Flush();
+			if (isAsync)
+			{
+				await outputStream.FlushAsync(ct);
+			}
+			else
+			{
+				outputStream.Flush();
+			}
```
piksel, line 619: Why is only this call branched? Shouldn't `await outputStream.FlushAsync(ct);` work fine for both cases? The default implementation of `FlushAsync` just calls `Flush` if not overriden.

src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
```diff
@@ -33,8 +33,18 @@ Please see https://github.com/icsharpcode/SharpZipLib/wiki/Release-1.3.3 for mor
...
+
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
+		<PackageReference Include="System.Memory" Version="4.5.4" />
+		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.2" />
+	</ItemGroup>
+
+	<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+		<PackageReference Include="System.Memory" Version="4.5.4" />
+		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.2" />
+	</ItemGroup>
```
piksel, line 45: There's been a core goal of keeping the library dependency free. But if it means keeping .NET 4.5 compatibility without sacrificing newer features and optimizations, maybe this ought to be reconsidered...

piksel (commented):
There should be some more tests with the async versions of the code, but otherwise it looks good. Great work!

src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
christophwille: https://devblogs.microsoft.com/dotnet/dotnet-framework-45-46-461-end-of-support/ maybe we should consider net462 minimum?

christophwille: We are netstandard2.0 on ICSharpCode.Decompiler - which means net472+ compat for the old full framework. In 2022, it should be ok to ditch anything older than that and use the polyfills.

src/ICSharpCode.SharpZipLib/Tar/TarBuffer.cs
nathan-c: The call to `WriteRecordAsync` (line 609) isnt branched as the `isAsync` parameter is passed down and it is handled internally.

There are 2 reasons to branch on `Flush` /  `FlushAsync`:
1. `FlushAsync` returns a `Task` and in the case of synchronous access there will be an unnecessary allocation here.
2. (More importantly) If the stream fully supports async usage but the caller explicitly calls the synchronous method it *might* be possible to [end up deadlocking](https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html) so I feel it is probably safer to branch.

src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
nathan-c: Is the conclusion to remove the `net45` target and keep the polyfill packages? Those packages provide `ValueTask` and ArrayPool / Span / Memory.  

I think the addition of `ValueTask` is important from a performance perspective as without it the new implementation would perform quite a bit worse than the existing one for sync calls, which is obviously not great.

It's probably possible to remove the dependency on `System.Memory` (without sacrificing too much) by going back to using plain old `byte[]` and pulling an implementation of `ArrayPool` into this code base (similar to the very basic `StringBuilderPool` I have already added). Though my preference would be to keep this package reference too if you are happy with that?

nathan-c:
Thanks for the feedback! I will create some more tests to cover the changes

nathan-c:
Some extra tests have now been added.

piksel (approved):

src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
piksel: Additionally, if we are dropping .NET 4.5 support, we can just stick to `netstandard2.0` as it's supported by 4.6.1-4.8:
https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0

I will make a new issue for this.

PR #750, opened 2022-04-28
fix(zip): 0 in zip64 local sizes using descriptors

### LINKED ISSUE #744: Different binary produced when producing ZIP to Azure Blob Stream
We found that our ZIP files were not being opened correctly on Macs, producing an error with `ditto`:
`ditto: Couldn't read pkzip signature.`

After further investigation, it seems that ZIP files produced to a local file (which can be opened on a Mac) are different to that produced to a stream that is persisting to Azure Blob Store. However, I also tested simply opening a file stream, and using the blob stream to persist it, which resulted in the same binary. I think this indicates that something with the zipping process is causing the problem.

From a binary comparison, it seems the differences are in the start/end of the files, though I am note sure which parts are significant, or show what could be going wrong.
On left is file persisted via Blob stream. Right is locally to disk.
Start:  
image  
End:  
image

You should be able to reproduce the issue with the following code. You can uncomment the various sections if you wish to write a local zip, or use a local file to upload to a blob store. Otherwise the uncommented code should produce a ZIP that would be invalid when opening on a Mac.
```csharp
using System;
using System.Net.Http;
using System.Threading.Tasks;
using Azure.Storage.Blobs;
using Azure.Storage.Blobs.Specialized;
using ICSharpCode.SharpZipLib.Zip;

internal class CreateZipFile
{
	public static void Main(string[] args)
	{
		MainAsync(args).GetAwaiter().GetResult();
	}

	private static async Task MainAsync(string[] args)
	{
		var httpClient = new HttpClient();

		var _blobServiceClient = new BlobServiceClient("DefaultEndpointsProtocol=https;AccountName=<replace_with_account_name>;AccountKey=<replace_with_a_key>;EndpointSuffix=core.windows.net");
		var _containerClient = _blobServiceClient.GetBlobContainerClient("artifacts");
		var blobName = System.IO.Path.ChangeExtension(Guid.NewGuid().ToString(), "zip");
		var blockClient = _containerClient.GetBlockBlobClient(blobName);
		var artifactUploadStream = blockClient.OpenWrite(overwrite: true);

		try
		{
			string[] filenames = new[] {
				"https://peach.blender.org/wp-content/uploads/bbb-splash.png"
			};

			// Writing an already produced ZIP to blob store results in matching binary.

			//var localFileMs = new MemoryStream(await System.IO.File.ReadAllBytesAsync("C:\\DevTools\\Debug\\o-local.zip"));
			//localFileMs.Position = 0;
			//await localFileMs.CopyToAsync(artifactUploadStream);
			//await artifactUploadStream.FlushAsync();
			//await artifactUploadStream.DisposeAsync();
			//await localFileMs.DisposeAsync();
			//return;

			// Local file writing produces expected ZIP
			//using var zipStream = new ZipOutputStream(System.IO.File.Create("C:\\DevTools\\Debug\\o-local.zip"))
			using var zipStream = new ZipOutputStream(artifactUploadStream)
			{
				IsStreamOwner = true
			};

			var downloadStream = await httpClient.GetStreamAsync("https://peach.blender.org/wp-content/uploads/bbb-splash.png");

			var zipFileEntry = new ZipEntry("big-buck-bunny.png")
			{
				CompressionMethod = CompressionMethod.Deflated,
				DateTime = new DateTime(2021, 7, 13, 4, 28, 44),
			};

			zipStream.PutNextEntry(zipFileEntry);
			await downloadStream.CopyToAsync(zipStream);
			zipStream.CloseEntry();
			zipStream.Finish();
			zipStream.Close();
		}
		catch (Exception ex)
		{
			Console.WriteLine("Exception during processing {0}", ex);
		}
	}
}
```

Not sure where else I can investigate further, or if this is more an issue with the Blob Block Stream, or even a mix of both?

Edit:
Here are the files to compare:
[remote.zip](https://github.com/icsharpcode/SharpZipLib/files/8452034/remote.zip)
[local.zip](https://github.com/icsharpcode/SharpZipLib/files/8452032/local.zip)

piksel:
In #736, the header tests were changed to allow for `0` being passed as the Zip64 sizes when testing the archive. That is actually the correct value as per the spec, and hence this updates the output of `ZipFormat.WriteLocalHeader` to write `0` when using descriptors together with Zip64.

Fixes #744

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -1145,12 +1145,12 @@ namespace ICSharpCode.SharpZipLib.Zip
 					if (localFlags.HasAny(GeneralBitFlags.Descriptor))
 					{
 						// These may be valid if patched later
-						if ((size > 0) && (size != entry.Size))
+						if ((size != 0) && (size != entry.Size))
 						{
 							throw new ZipException("Size invalid for descriptor");
 						}
 
-						if ((compressedSize > 0) && (compressedSize != entry.CompressedSize))
+						if ((compressedSize != 0) && (compressedSize != entry.CompressedSize))
 						{
 							throw new ZipException("Compressed size invalid for descriptor");
 						}
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFormat.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFormat.cs
@@ -114,8 +114,9 @@ namespace ICSharpCode.SharpZipLib.Zip
 				}
 				else
 				{
-					ed.AddLeLong(-1);
-					ed.AddLeLong(-1);
+					// If the sizes are stored in the descriptor, the local Zip64 sizes should be 0
+					ed.AddLeLong(0);
+					ed.AddLeLong(0);
 				}
 				ed.AddNewEntry(1);
 


piksel:
@carlreid no release date, but I should have the time to do so pretty soon. sorry that I cannot be more specific.

carlreid:
> sorry that I cannot be more specific.

No stress! We can wait a little more

PR #755, opened 2022-05-24
prepare for v1.4.0

piksel:
this branch contains changes needed for the upcoming release, mainly related to .NET 6

The new target frameworks will be:
- .NET Standard 2.0 (for .NET framework 4.6.2+)
- .NET Standard 2.1 (general support for all but .NET framework)
- .NET 6.0 (new LTS)

the main issue stems from the fact that the AES auth code reading is handled inside a class inheriting from `CryptoStream`. For that to work correctly, the reads cannot be bypassed by more efficient code (which is what is happening in the .NET 6 version of `CryptoStream`).
In fact, the entire `ZIPAESStream` should probably just be removed altogether, since it seems to just be a work around for having the auth code appended to the input stream. this can be handled much easier in the TransformLastBlock method of ZipAESTransform.

that approach is now used by async reads on .NET 6.0, but others still use the legacy code. this was done to do as little as possible to allow SharpZipLib to work with .NET 6, but it really ought to be replaced for all usages.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -18,11 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-2019, macos-latest]
-        target: [netstandard2.0, netstandard2.1]
-        include:
-          - os: windows-2019
-            target: net45
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        target: [netstandard2.0, netstandard2.1, net6.0]
     env:
       LIB_PROJ: src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
     steps:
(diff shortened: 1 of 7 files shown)


tonyliving:
Can I get it from nuget?

piksel:
@tonyliving not until it has been released, no. If you want to try it out you can download the build artifact for this PR:
https://github.com/icsharpcode/SharpZipLib/suites/6639150865/artifacts/250580285

carlreid:
No worries, take it easy 😊

carlreid:
It seems the artifact from the previous build has expired, is it possible you could re-trigger the build / release this version?

Was just about to host the pre-release package so we could make use of the fix https://github.com/icsharpcode/SharpZipLib/pull/750

piksel:
@carlreid I fixed the conflict and there should be a new artifact to download. Also, I fixed the last PR that was blocking release, so it *should* be ready by the weekend.

carlreid:
Wonderful! Thank you 😊

In which case, I'll wait until next week to see if you managed to make it through with a release before trying to host the artifact.

PR #765, opened 2022-08-10
Fix security weakness for compilation in .NET 7

yihezkel:
The Rfc2898DeriveBytes constructor used is called out as insecure in NET 7, because it’s insecure to use defaults for the number of iterations and hashing algorithm. The fix is to pass those (secure) values in the versions of .NET that support it.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs
+++ b/src/ICSharpCode.SharpZipLib/Encryption/ZipAESTransform.cs
@@ -76,7 +76,11 @@ namespace ICSharpCode.SharpZipLib.Encryption
 			_encrPos = ENCRYPT_BLOCK;
 
 			// Performs the equivalent of derive_key in Dr Brian Gladman's pwd2key.c
+#if NET472_OR_GREATER || NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_0_OR_GREATER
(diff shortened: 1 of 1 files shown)


yihezkel:
From https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rfc2898derivebytes.-ctor?view=net-7.0#system-security-cryptography-rfc2898derivebytes-ctor(system-string-system-int32-system-int32)

"The default hash algorithm and iteration counts in Rfc2898DeriveBytes constructors are outdated and insecure. Use a constructor that accepts the hash algorithm and the number of iterations.", DiagnosticId="SYSLIB0041", UrlFormat="https://aka.ms/dotnet-warnings/{0}"

piksel:
Well, the "insecure default" value for HashAlgorithmName is SHA1, which is what we have to use for the format. So, this would not increase security in any way, but it _would_ replace a deprecated version with a current one. And also, it's generally better to be explicit rather than relying on defaults.

yihezkel:
Exactly. The intent here is to be explicit, which is a better approach, but also satisfies .NET 7 compilation requirements :-)

If in the future there's interest to upgrade the hashing algorithm, this change should make that a tiny bit easier.

piksel (approved):
LGTM!

PR #766, opened 2022-08-15
fix: replace uses of obsolete method `RNGCryptoServiceProvider`

yihezkel:
Replace insecure obsolete method (new RNGCryptoServiceProvider()) with RandomNumberGenerator.Create() in PkzipClassic, ZipFile and ZipOutputStream.


yihezkel:
See https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider?view=net-6.0 and https://github.com/dotnet/runtime/issues/40169

piksel (approved):
Yeah, I actually did this in the `v1.4` branch, but this is more explicit and thorough. Thanks!

PR #778, opened 2022-09-22
fix(zip): cleanup/fix of StringCodec

### LINKED ISSUE #776: No way to set StringCodec on ZipInputStream & ZipOutputStream
[Obsolete("Use ZipFile/Zip*Stream StringCodec instead")]
I got this from the latest version. But ZipInputStream & ZipOutputStream do not provide any StingCodec property and the only one which might be useful constructor is internal
image

### Expected behavior
Should allow us to set the StringCodec if ZipStrings.CodePage is obsolete

### Actual behavior
No way except the obsolete way to set the StringCodec

### Version of SharpZipLib
1.4.0

### LINKED ISSUE #777: Files and folders with non-english characters names are have gibberish names after unpacking
### Steps to reproduce
1. Pack "слово.txt" by any archivator, ex. Bandizip
2. See that any programm unpacks it correctly (ex. windows explorer or same program you packed it)
3. Unpack it via `new FastZip().ExtractZip()1

### Expected behavior
files and folders have same name as when they was packed

### Actual behavior
files and folders name is like ����� �ணࠬ��

### Version of SharpZipLib
1.4.0

### Tryed actions:
```cs
ZipStrings.CodePage = 866;    // No data is available for encoding 866. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method.
ZipStrings.CodePage = 1251;   // same error
ZipStrings.CodePage = 65001;  // gibberish
ZipStrings.UseUnicode = true; // gibberish

new FastZip { 
  EntryFactory = new ZipEntryFactory { 
    IsUnicodeText = true 
  } 
}.ExtractZip(filepath, (TempFolder? "Temp\\" : ""), null);
```

(instead of just 
```cs
new FastZip ().ExtractZip(filepath, (TempFolder? "Temp\\" : ""), null);
//not works. 
``` 

`IsUnicodeText = true` gives same result as `IsUnicodeText = false`

piksel:
This PR fixes some issues that were present in #592 but wasn't detected before the release of v1.4.0.

In general, it goes a bit further in trying to make the string encoding easy and predictable to use.
It also has tests for the most common usages of the API to make sure it both works and feels straightforward to use.
The changes in the API are slightly breaking from v1.4.0, but only where the API could be misused in confusing ways. It should not require any minor version bump, but should be released ASAP to limit the effect.

Fixes #777.
Fixes #776.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/DeflaterOutputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/DeflaterOutputStream.cs
@@ -205,7 +205,12 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 		protected byte[] AESAuthCode;
 
 		/// <inheritdoc cref="StringCodec.ZipCryptoEncoding"/>
-		public Encoding ZipCryptoEncoding { get; set; } = StringCodec.DefaultZipCryptoEncoding;
+		public Encoding ZipCryptoEncoding {
+			get => _stringCodec.ZipCryptoEncoding;
+			set {
+				_stringCodec = _stringCodec.WithZipCryptoEncoding(value);
+			} 
+		}
 
 		/// <summary>
 		/// Encrypt a block of data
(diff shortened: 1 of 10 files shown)


filipnavara:
Any news on this one? It's blocking us from updating to the latest NuGet.

piksel:
Sorry, I have been busy with work. There are some inconsistencies in the API that I need to look over, but it should mostly be done.

PR #787, opened 2022-10-19
fix: use ConfigureAwait(false) on every await

### LINKED ISSUE #786: Regression 1.3.3 => 1.4.0: Deadlock when using TarInputStream.GetNextEntry() in WPF application
### Steps to reproduce

This is part of a large WPF application (.NET Framework 4.8), I have to build a small reproduction. I just want to mention this issue in case someone has the same problem.

### Expected behavior
`TarInputStream.GetNextEntry()` should not deadlock when used on the main thread.

### Actual behavior
`TarInputStream.GetNextEntry()` blocks when called on the main thread of an WPF application. In the PR https://github.com/icsharpcode/SharpZipLib/pull/746 for the async streams I see a lot of `.GetAwaiter().GetResult()`. This is not a recommended practice, see https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#warning-deadlocks

### Version of SharpZipLib
1.4.0

flensrocker:
See https://devblogs.microsoft.com/dotnet/configureawait-faq/

General purpose libraries should always use `.CongfigureAwait(false)` to avoid deadlocks. It may not prevent from every deadlock but is after all a "best practice".

I also added a `.globalconfig` to warn if `ConfigureAwait` calls are missing in future changes.

Fixes #786 

### DIFF
--- a/.globalconfig
+++ b/.globalconfig
@@ -1,0 +1,3 @@
+is_global = true
+global_level = 1
+dotnet_diagnostic.CA2007.severity = warning
--- a/benchmark/.globalconfig
+++ b/benchmark/.globalconfig
@@ -1,0 +1,3 @@
+is_global = true
+global_level = 2
+dotnet_diagnostic.CA2007.severity = none
--- a/src/ICSharpCode.SharpZipLib/Core/ByteOrderUtils.cs
+++ b/src/ICSharpCode.SharpZipLib/Core/ByteOrderUtils.cs
@@ -80,7 +80,7 @@ namespace ICSharpCode.SharpZipLib.Core
 		/// <inheritdoc cref="WriteLEShort"/>
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		public static async Task WriteLEShortAsync(this Stream stream, int value, CT ct) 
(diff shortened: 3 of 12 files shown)


piksel:
This is absolutely better overall, but it does only cover up the real bug. The "point" of the TarInputStream async implementation is that it's supposed to run entirely sync when being called with `isAsync = false` ("This async method lacks ‘await’ operators and will run synchronously").

This *should* mean that a deadlock would never occur.

Adding the `ConfigureAwait(false)` to all the await calls will increase the performance though, and is a great catch (far too long since I did async library code, I had actually forgot about it), but it will only bypass the bug here by using another thread.

I will kill the `Fixes`/auto-close though and create a new issue for it, and get this released ASAP.

piksel (approved):

flensrocker:
If I find time I will take a look at the generated IL (never had that done before) to see, if `GetNextEntry` will really be sync all the way. I don't know what the compiler does when it comes to `SkipToNextEntry` etc.. The `await`s are a bit nested until it comes to the `TarBuffer.ReadRecordAsync` where the `isAsync` is evaluated.

Thank you for merging!

PR #791, opened 2022-10-20
fix(tar): use sync codepath on sync methods

### LINKED ISSUE #788: TarInputStream does not run synchronously when used non-async
This is a follow up from #786.

`TarInputStream.GetNextEntry()` seems to use the thread pool to run "asynchronously", even though it should be fully sync.
The async implementation takes an `isAsync` argument that should make it run totally synchronous when passed with `false`.

As of #787, deadlocks are avoided using `.ConfigureAwait(false)`, but there is still a bug somewhere that caused this to happen in the first place.

flensrocker:

Fixes #788


piksel:
Ehm, yeah, that was much more obvious than I thought. Probably should have started looking there... 😁

flensrocker:
I looked at those lines more than once... and didn't see it.

I started to use enums instead of booleans for such switches because they are more readable.

PR #818, opened 2022-12-20
Add support for DEFLATE64 algorithm when extracting from zip archive

elitzamarinova:
Add support for DEFLATE64 algorithm when extracting from zip archive:

The content of the new classes is borrowed from DEFLATE64 support implementation for DotNetZip which on its part contains modified code from the .NET Core Libraries (CoreFX and System.IO.Compression/DeflateManaged) where deflate64 decompression is implemented.

Other changes are implemented in order to support forward-only stream.

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
@@ -109,6 +109,24 @@ namespace ICSharpCode.SharpZipLib.Zip.Compression.Streams
 		}
 
 		/// <summary>
+		/// Resize input buffer according to a specific array size
+		/// </summary>
+		/// <param name="bufferSize"></param>		
+		public void ResizeBuffer(int bufferSize)
+		{
+			if (available == 0)
+			{
+				int oldSize = rawData.Length;
+				byte[] resized = rawData;
+				Array.Resize(ref resized, bufferSize);
+				rawData = resized;
+				rawLength = rawData.Length;
+				clearText = rawData;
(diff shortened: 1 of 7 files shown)


src/ICSharpCode.SharpZipLib/Zip/Compression/Streams/InflaterInputStream.cs
```diff
@@ -108,6 +108,24 @@ public void SetInflaterInput(Inflater inflater)
...
+		/// Resize input buffer according to a specific array size
+		/// </summary>
+		/// <param name="bufferSize"></param>		
+		public void ResizeBuffer(int bufferSize)
+		{
+			if (available == 0)
+			{
+				int oldSize = rawData.Length;
+				byte[] resized = rawData;
+				Array.Resize(ref resized, bufferSize);
```
christophwille, line 121: You are aware that is a costly copy operation?

elitzamarinova: Yes, I am. But in this particular case I don't think that it is significant.

piksel:
This currently fails 4 tests, and I am not too eager in taking in a whole new Deflate implementation (1800 LoC 😓).   

I did some PoC testing in https://github.com/icsharpcode/SharpZipLib/commit/c9589cf493b8fe4f3146d2b72fc2847b068ece45 to see if it could be implemented by allowing the window and max distance to be customized in the existing Deflater.
It still needs more testing and a Inflator support, but it looks promising...

elitzamarinova:
Yes, looks promising.
I took a look today at the failings in the tests, seems like the ZipEntry in the ZipInputStream got null somehow.

PR #840, opened 2023-08-03
Fix ZipFile constructor breaking change

### LINKED ISSUE #839: SharpZipLib v1.4 introduced breaking change
### Describe the bug

ZipFile class constructor prior v1.4 had this constructor:
`public ZipFile(string name)`

In v1.4 optional parameter `StringCodec` was added which causing `Method not found` exception in other libraries built with older versions of SharpZipLib. It's a breaking change, instead of adding optional parameter you should create 2 separate constructors.

Updating SharpZipLib to v1.4 breaks NPOI: https://github.com/nissl-lab/npoi/issues/1149

### Reproduction Code

_No response_

### Steps to reproduce

described

### Expected behavior

No breaking change should be introduced

### Operating System

Windows, macOS, Linux

### Framework Version

.NET 7, .NET 6, .NET 5, .NET Core v3 and earlier, .NET Framework 4.x

### Tags

ZIP

### Additional context

_No response_

Bykiev:
This PR fixes #839

### DIFF
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -391,6 +391,19 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// Opens a Zip file with the given name for reading.
 		/// </summary>
 		/// <param name="name">The name of the file to open.</param>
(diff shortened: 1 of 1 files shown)


piksel (approved):
Thanks. This wasn't an intended break.

PR #843, opened 2023-08-10
Improve zip file iteration performance

lahma:
When traversing large zip files I noticed that SharpZipLib was allocating quite a lot of memory. So I found out that by pooling of costly `Inflator` instances a lot of that can be rectified. There's now new `PooledInflator` type that can be used by the library itself and is known to be safe to pool (it's rented from pool or is otherwise instance created by the library itself).

I also changed ZipFile to follow the common `public struct` enumerator pattern found in BCL data structures that removes unnecessary allocations 

Benchmark test case updated and now targets oldest supported runtimes (.NET 4.6.2 and .NET 6.0). Unsupported runtimes are not installed by default.

# Results

```

BenchmarkDotNet v0.13.7, Windows 11 (10.0.23516.1000)
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK 7.0.400
  [Host]     : .NET 6.0.21 (6.0.2123.36311), X64 RyuJIT AVX2
  Job-ERDPDJ : .NET 6.0.21 (6.0.2123.36311), X64 RyuJIT AVX2
  Job-HFJNUX : .NET Framework 4.8.1 (4.8.9179.0), X64 RyuJIT VectorSize=256


```

## ICSharpCode.SharpZipLib.Benchmark.Zip.ZipInputStream

| **Diff**|Method|Toolchain|Mean|Error|Allocated|
|------- |-------|-------|-------:|-------|-------:|
| Old |ReadZipInputStream|.NET 6.0|151.4 ms|0.60 ms|104.54 KB|
| **New** |	| **.NET 6.0** | **151.7 ms (0%)** | **0.84 ms** | **72.15 KB (-31%)** |
| Old |ReadZipInputStream|.NET Framework 4.6.2|150.2 ms|0.50 ms|106.05 KB|
| **New** |	| **.NET Framework 4.6.2** | **150.8 ms (0%)** | **0.70 ms** | **74 KB (-30%)** |
| Old |ReadLargeZipFile|.NET 6.0|1,337.3 ms|8.18 ms|2412174.27 KB|
| **New** |	| **.NET 6.0** | **1,292.3 ms (-3%)** | **9.08 ms** | **548513.15 KB (-77%)** |
| Old |ReadLargeZipFile|.NET Framework 4.6.2|1,499.4 ms|11.41 ms|2421273.81 KB|
| **New** |	| **.NET Framework 4.6.2** | **1,452.4 ms (-3%)** | **3.35 ms** | **556703.89 KB (-77%)** |

### DIFF
--- a/src/ICSharpCode.SharpZipLib/GZip/GzipInputStream.cs
+++ b/src/ICSharpCode.SharpZipLib/GZip/GzipInputStream.cs
@@ -4,6 +4,7 @@ using ICSharpCode.SharpZipLib.Zip.Compression.Streams;
 using System;
 using System.IO;
 using System.Text;
+using ICSharpCode.SharpZipLib.Core;
 
 namespace ICSharpCode.SharpZipLib.GZip
 {
@@ -82,7 +83,7 @@ namespace ICSharpCode.SharpZipLib.GZip
 		/// Size of the buffer to use
 		/// </param>
 		public GZipInputStream(Stream baseInputStream, int size)
-			: base(baseInputStream, new Inflater(true), size)
+			: base(baseInputStream, InflaterPool.Instance.Rent(true), size)
 		{
 		}
 
--- a/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
+++ b/src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
@@ -314,7 +314,7 @@ namespace ICSharpCode.SharpZipLib.Zip
 	/// }
 	/// </code>
 	/// </example>
-	public class ZipFile : IEnumerable, IDisposable
+	public class ZipFile : IEnumerable<ZipEntry>, IDisposable
 	{
 		#region KeyHandling
 
@@ -810,7 +810,31 @@ namespace ICSharpCode.SharpZipLib.Zip
 		/// <exception cref="ObjectDisposedException">
 		/// The Zip file has been closed.
 		/// </exception>
-		public IEnumerator GetEnumerator()
+		IEnumerator IEnumerable.GetEnumerator()
+		{
+			return GetEnumerator();
+		}
+
+		/// <summary>
+		/// Gets an enumerator for the Zip entries in this Zip file.
+		/// </summary>
+		/// <returns>Returns an <see cref="IEnumerator"/> for this archive.</returns>
+		/// <exception cref="ObjectDisposedException">
+		/// The Zip file has been closed.
+		/// </exception>
+		IEnumerator<ZipEntry> IEnumerable<ZipEntry>.GetEnumerator()
+		{
+			return GetEnumerator();
+		}
+
+		/// <summary>
+		/// Gets an enumerator for the Zip entries in this Zip file.
+		/// </summary>
+		/// <returns>Returns an <see cref="IEnumerator"/> for this archive.</returns>
+		/// <exception cref="ObjectDisposedException">
+		/// The Zip file has been closed.
+		/// </exception>
+		public ZipEntryEnumerator GetEnumerator()
 		{
 			if (isDisposed_)
 			{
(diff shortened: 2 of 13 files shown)


src/ICSharpCode.SharpZipLib/Zip/ZipFile.cs
```diff
@@ -954,7 +978,7 @@ public Stream GetInputStream(long entryIndex)
 
 				case CompressionMethod.Deflated:
 					// No need to worry about ownership and closing as underlying stream close does nothing.
-					result = new InflaterInputStream(result, new Inflater(true));
+					result = new InflaterInputStream(result, InflaterPool.Instance.Rent(true));
```
piksel, line 981: Again, to avoid surprise breakage on low-resource setups, perhaps make this depend on a public property, like `EnableInflatorPooling`?

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -88,7 +89,7 @@ public class ZipInputStream : InflaterInputStream
 		/// </summary>
 		/// <param name="baseInputStream">The underlying <see cref="Stream"/> providing data.</param>
 		public ZipInputStream(Stream baseInputStream)
-			: base(baseInputStream, new Inflater(true))
+			: base(baseInputStream, InflaterPool.Instance.Rent(true))
```
piksel, line 92: Again, named constructor instead of default.

src/ICSharpCode.SharpZipLib/Zip/ZipInputStream.cs
```diff
@@ -99,7 +100,7 @@ public ZipInputStream(Stream baseInputStream)
 		/// <param name="baseInputStream">The underlying <see cref="Stream"/> providing data.</param>
 		/// <param name="bufferSize">Size of the buffer.</param>
 		public ZipInputStream(Stream baseInputStream, int bufferSize)
-			: base(baseInputStream, new Inflater(true), bufferSize)
+			: base(baseInputStream, InflaterPool.Instance.Rent(true), bufferSize)
```
piksel, line 103: Again, named constructor instead of default.

piksel (changes requested):
I think the pooling looks sound in general, but I think it should be opt-in due to how it might affect more resource constrained setups.

src/ICSharpCode.SharpZipLib/GZip/GzipInputStream.cs
```diff
@@ -82,7 +83,7 @@ public GZipInputStream(Stream baseInputStream)
 		/// Size of the buffer to use
 		/// </param>
 		public GZipInputStream(Stream baseInputStream, int size)
-			: base(baseInputStream, new Inflater(true), size)
+			: base(baseInputStream, InflaterPool.Instance.Rent(true), size)
```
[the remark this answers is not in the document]
lahma: Factory methods are a bit problematic, as I for example am using `ZipFile.GetInputStream(long entryIndex)` which makes it really hard to hint which kind of constructor should be used. I'll propose lowing the default pool size and allowing to disable it.

lahma:
Thank you for the review. As I mentioned in comment, the ZipFile scenario is hard to use when you cannot control to internal input stream creation. I created a fix commit that proposes to have `InflaterPool.Size` which is only public member and allows to disable the pooling by supplying a value of 0 or less.

What do you think about the approach? I would also gladly any commits from you into this part shaping the solution towards your liking, but also hoping that the benchmark's `ZipFile` API usage would be covered.

piksel:
I think that would be a good compromise. I am not sure about making the pool public though... Perhaps there should be a public singleton that maps to the internals of the pool.

That way we have a clear consumer API for global library tweaks... Or perhaps I am just overthinking it.

The initial value would need to be 0 though, as the goal is to not accidently break anything (but again, maybe I am making this into a larger issue than it needs to be..)

lahma:
It's always good to keep downstream consumers happy! 😉  I've force pushed a proposal of adding root level static class for global options, `SharpZipLibOptions`, which has now sole property of `InflaterPoolSize` which defaults to 0. Benchmark updated to set the value to 4 (allocations still down 70%).

piksel (approved):
Looks great! The performance improvements are nice and this should not interfere with any existing usage (afaict).

rizucoorg97 (approved):

PR #872, opened 2024-07-18
Add net462 targeting

### LINKED ISSUE #853: .NET Framework 4.6.2 Support?
Hello,

I noticed that as of v1.4.0 support for .NET 4.5 Framework has been dropped in favor of .NET 6. This has also been mentioned on issue #804, and the accepted solution was the fact that .NET 4.8 is supported through the `netstandard2.0` target.

However, when it comes to older frameworks such as .NET 4.6.1/4.6.2, even though they do support .NET Standard 2.0, when compiling a project that references netstandard libraries, it generates an additional 96 assemblies it the build output, which is quite annoying.

Because of that, many nuget packages also provide a `net461` or `net462` target, that allows them to be referenced without producing those additional assemblies. No code changes are required, since all the NETStandard APIs are present in those frameworks.

Therefore I would like to request that framework to be included as a build target in this project. Since .NET 4.6.1 has recently come to end of support, maybe you could add `net462` as a target to this library.

Anyway, thanks for your attention.

### LINKED ISSUE #863: Add support for .NET Framework 4.6.1 to target frameworks.
### Is your feature request related to a problem? Please describe.

We've been using prior versions of SharpZipLib with .NET Framework and once upgrading, we are running into issues at compile time referencing .NET Standard 2.0. For various reasons, we're unable to upgrade our .NET Framework version up to 4.7.2+. Per Microsoft:
> While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher. (https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0)

### Describe the solution you'd like

Add net461 to this [line](https://github.com/icsharpcode/SharpZipLib/blob/ff2d7c30bdb2474d507f001bc555405e9f02a0bb/src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj#L4). We are not sure if there's more that would need to happen as we are not familiar with this code base.

### Describe alternatives you've considered

The only other solution that we would be able to do would be to compile this ourselves internally.

### Tags

_No response_

### Additional context

_No response_

lahma:
It's not that much of a maintenance burden and will remove some compatibility issues for folks still on full framework. SharpZipLib 2.0 could drop support if deemed so.

Also had to update testing dependencies to get NET 8 support.

fixes #853
fixes #863

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -19,17 +19,17 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        target: [netstandard2.0, netstandard2.1, net6.0]
+        target: [netstandard2.0, netstandard2.1, net6.0, net462]
     env:
       LIB_PROJ: src/ICSharpCode.SharpZipLib/ICSharpCode.SharpZipLib.csproj
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         ref: ${{ github.events.inputs.tag }}
         fetch-depth: 0
 
     - name: Setup .NET
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v4
       with:
         dotnet-version: '6.0.x'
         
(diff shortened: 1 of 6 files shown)


lahma:
@piksel can you please take a look, this PR should fix GH Actions pipelines and adds the requested net462 target which should be quite safe addition.

